PR #2, opened 2016-04-06
Change database config to match other apps

brenetic:
- Refactor database config to be more in line with other govuk apps

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,25 +1,22 @@
-development:
+default: &default
   adapter: postgresql
-  database: local_links_manager_development
+  encoding: unicode
   template: template0
-  encoding: utf8
-  variables:
-    work_mem: 10MB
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -1,25 +1,22 @@
...
-    work_mem: 1MB
+  <<: *default
+  database: local-links-manager_test<%= ENV['TEST_ENV_NUMBER'] %>
 
 cucumber:
   <<: *test
 
 production:
+  <<: *default
   url: <%= ENV['DATABASE_URL'] %>
```
sihugh, line 22: I haven't seen this approach before.  How does this work with the env variables hash set in puppet?  I notice that content-tagger uses username/password/database. (Actually commenting on the line below: `url: <%= ENV['DATABASE_URL'] %>` )

sihugh:
:+1:

PR #4, opened 2016-04-19
Import local authorities

sihugh:
Added LocalAuthority model

Added rake task to import local authorities from MapIt
Added rake task to import slugs for each authority from [`authorities.json`](https://github.com/alphagov/frontend/blob/master/lib/data/authorities.json) in Frontend
Added rake task to import homepage URLs from the local authority contacts [CSV](http://local.direct.gov.uk/Data/local_authority_contact_details.csv) file on http://local.direct.gov.uk/Data 

Added UI to display the imported list of local authorities

screen shot 2016-04-27 at 17 04 06


spec/rails_helper.rb
```diff
@@ -58,4 +58,12 @@
...
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+end
+
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
```
h-lame: This sort of thing should go in a file in `spec/support/` as it makes it easier to upgrade rspec-rails later without worrying too much about incorporating changes to the `rails_helper` that new versions of rspec-rails would bring.

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def get_mapit_authorities(url)
+    raw_authorities = get_json(url)
+    raw_authorities.values.map { |raw_authority|
+      get_la_struct(raw_authority)
+    }
+  end
+
+  def get_la_struct(raw_authority)
+    authority = OpenStruct.new
```
davidbasalla: Are you using OpenStruct for a specific reason here over normal Struct? Afaik people have ['opinions'](http://palexander.posthaven.com/ruby-data-object-comparison-or-why-you-should-never-ever-use-openstruct) on using OpenStruct in production code. On the other hand, I know it's being used in various GDS apps so might be fine?

Edit: What @h-lame said :)

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,63 @@
...
+    authority.name = raw_authority["name"]
+    authority.snac = raw_authority["codes"]["ons"]
+    authority.gss = raw_authority["codes"]["gss"]
+    authority.tier = identify_tier(raw_authority["type"])
+    authority
+  end
```
h-lame: We should avoid `OpenStruct` if we can (see: http://blog.honeybadger.io/how-openstruct-and-hashes-can-kill-performance/).  In this case we know what the attributes are, so we could define a `Struct` instead:

```
LAStruct = Struct.new(:name, :snac, :gss, :tier)
def get_la_struct(raw_authority)
  authority = LAStruct.new
  authority.name = raw_authority["name"]
  authority.snac = raw_authority["codes"]["ons"]
  authority.gss = raw_authority["codes"]["gss"]
  authority.tier = identify_tier(raw_authority["type"])
  authority
end
```

But! I'm not sure there's much benefit in having an object here at all when all we're going to do with it is immediately turn it into a new `LocalAuthority` or use it to update an existing one.  It's still useful to translate from the deeply nested hash we get from mapit though, but maybe a simple hash would suffice?

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,63 @@
+require 'ostruct'
+
+class LocalAuthoritiesImporter
+  include JsonDownloader
+
+  LOCAL_AUTHORITY_TYPES = "DIS,LBO,MTD,UTA,COI,LGD"
```
h-lame: This doesn't include `CTY`, so we won't be getting Counties.

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def identify_tier(area_type)
+    case area_type
+    when 'DIS' then 'district'
+    when 'CTY' then 'county'
+    when 'LBO', 'LDG', 'MTD', 'UTA' then 'unitary'
+    else
+      'county'
+    end
+  end
```
h-lame: This method doesn't make mention of `COI` (Isle of Scilly) so it'll fall through to 'county', but that might not be correct.  It might be worth making sure we only have one place that defines which mapit area types we care about to avoid cases where we do different things to different sets of types.  So combining the constant `LOCAL_AUTHORITY_TYPES` above and the values used in this method; perhaps a constant Hash instance mapping types to "names" with a default value for unknown keys?

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def self.import_from_mapit
+    new.authorities_from_mapit
+  end
+
+  def authorities_from_mapit
+    url = "#{MAPIT_BASE_URL}areas/#{LOCAL_AUTHORITY_TYPES}"
+    mapit_las = get_mapit_authorities(url)
+
+    mapit_las.each do |mapit_la|
```
davidbasalla: Possibly nitpicking: Could you rename the `get_mapit_authorities` method to just `mapit_authorities` and then call `each` on that directly? I'm also wondering whether the `url` variable could become a class constant/method, so that you don't have to pass it as a parameter.

lib/local_authorities_slug_importer.rb
```diff
@@ -0,0 +1,42 @@
...
+    @authorities = get_json(AUTHORITIES_URL)
+  end
+
+  def find_slug_from_gss(gss)
+    las = authorities.select do |_slug, codes|
+      codes["gss"] == gss.to_s
+    end
+
+    las.keys.first
+  end
```
h-lame: In terms of perf we could use [`#detect`](http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-detect) instead of [`#select`](http://ruby-doc.org/core-2.3.0/Hash.html#method-i-select) - as `detect` will stop as soon as it finds a match, rather than collecting all of them.  Note that `detect` on a hash comes from `Enumerable` and so it returns an array of `[key, value]` rather than a hash.  This means we'd also have to change the final line to return the correct value.

h-lame: We could also mutate the data we get from the `authorities.json` to turn it into the exact form we need: if it was a hash of `{gss => slug}` that would make `find_slug_from_gss` much simpler.  It should also be faster as we mutate the data once (expensive) but then each search is v. fast, rather than now where we don't mutate, but do search many times and the search is slow(ish).

lib/local_authorities_url_importer.rb
```diff
@@ -0,0 +1,93 @@
...
+
+
+  def self.import_urls
+    fh = fetch_data
+    begin
+      new(fh).import
+    ensure
+      fh.close
+    end
+  end
```
h-lame: We can rejig `fetch_http_to_file` to take a block and avoid having to write this `begin...ensure...end`.  If in `fetch_http_to_file` we used [`Tempfile.open`](http://ruby-doc.org/stdlib-2.3.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-open) instead of `.new` we can pass it a block and it will deal with closing and unlinking the file automatically.  Something like:

``` ruby
def self.import_urls
  fetch_data do |fh|
    new(fh).import
  end
end

def self.fetch_data(&block)
  block.call fetch_http_to_file(CONTACTS_LIST_URL)
end

def self.fetch_http_to_file(url, &block)
  Tempfile.open(%w(local_authority_data csv) do |fh|
    # existing code
    block.call(fh)
  end
end
```

It doesn't really save us much typing, but it is a wee bit more idiomatic for rubyists to use blocks (especially when dealing with files) to allow the ruby stdlib to deal with cleanup for us.

lib/local_authorities_url_importer.rb
```diff
@@ -0,0 +1,93 @@
...
+private
+
+  def process_row(row)
+    return if row['SNAC Code'].blank?
+    authority = LocalAuthority.where(snac: row['SNAC Code']).first
+    unless authority
+      Rails.logger.warn "#{self.class}: failed to find LocalAuthority with SNAC #{row['SNAC Code']}"
+      return
+    end
+    #authority.name = decode_broken_entities( row['Name'] )
```
h-lame: Do we not need this anymore?  Or was it commented out in publisher too?  We could probably just remove it if we don't need it (the method and the line invoking it).

app/views/layouts/application.html.erb
```diff
@@ -3,4 +3,8 @@
   <%= javascript_include_tag 'application' %>
 <% end %>
 
+<% content_for :navbar_items do %>
+	<li><a href="/local-authorities">Local Authorities</a></li>
+<% end %>
```
h-lame: This probably wants to be 2-spaces, not 1-tab.  Rubocop doesn't run on erb files, so the linter won't have caught it.

lib/local_authorities_url_importer.rb
sihugh: The Trello card considers using the names from the CSV which was why it was in there, but I'll remove it for now.  It's easy enough to add in later if necessary

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,72 @@
+require 'json_downloader'
+
+class LocalAuthoritiesImporter
```
jennyd: It would be nice to start out with an `import` module - this PR adds 3 importers and we know we need to write more (for services, interactions and links), so we can keep them clearly separated from other lib code. [Transition does this](https://github.com/alphagov/transition/tree/master/lib/transition/import) and it similarly imports data from various sources.

lib/local_authorities_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+    "UTA" => UNITARY,
+  }
+
+  def self.import_from_mapit
+    new.authorities_from_mapit
+  end
+
+  def authorities_from_mapit
+    url = "#{MAPIT_BASE_URL}areas/#{local_authority_types}"
+    mapit_las = get_mapit_authorities(url)
```
jennyd: We'd usually use gds-api-adapters for internal API calls - there's a [method there which makes this call](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/mapit.rb#L11-L13) (although its name suggests that that endpoint only accepts a single type).

sihugh: Thanks

lib/local_authorities_importer.rb
sihugh: I'd taken a lead from the importers in Publisher which aren't done in this way.  I'm happy to change though as we are likely to have more Rake tasks etc. besides imports in future.

config/initializers/mapit.rb
```diff
@@ -0,0 +1 @@
+MAPIT_BASE_URL = "#{Plek.find('mapit')}/"
```
jennyd: Using gds-api-adapters means that we can instantiate the adapter here, instead of just getting the URL. Some apps (eg [smart-answers](https://github.com/alphagov/smart-answers/blob/master/lib/services.rb), [publisher](https://github.com/alphagov/publisher/blob/master/lib/services.rb), [collections-publisher](https://github.com/alphagov/collections-publisher/blob/master/lib/services.rb)) seem to be using a `Services` module now instead of an initializer for this, which we could do (although we only have mapit for now).

lib/tasks/local_authorities.rake
```diff
@@ -0,0 +1,6 @@
+namespace :local_authorities do
```
jennyd: Similarly to the `import` module comment above, we could use an `import` namespace above `local_authorities` so that these are grouped with all the future ones.

lib/tasks/local_authorities.rake
```diff
@@ -0,0 +1,6 @@
...
+  desc "Import local authority names, codes and tiers from MapIt"
+  task import_codes: :environment do
```
jennyd: This task is importing more than just codes - it's the full list of all LAs and determines which ones exist for us. Although we import other data about LAs from other sources, those tasks don't create new authorities, they only update particular attributes. Can we find a name which makes it clearer that this is the main one, maybe `import_list`?

The description could also make it clearer that this is the one which does the creating by starting "Import the list of..."  rather than just "Import...", and then the descriptions for the slug and homepage URL tasks could be "Add [slugs/homepage URLs] to local authorities from [source]" which would also clarify the relationships between the tasks.

spec/lib/local_authorities_importer_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    stub_mapit_request(updated_name_type_and_ons)
+
+    LocalAuthoritiesImporter.import_from_mapit
+
+    expect(LocalAuthority.count).to eq(8)
+
+    la = LocalAuthority.find_by(gss: 'S12000033')
+
+    expect(la.name).to eq("A Different Council")
+    expect(la.snac).to_not eq("XXXX")
```
jennyd: Why don't we want to update the SNAC? They're unlikely to change in practice since they aren't being issued any more, but we might need to correct a SNAC in mapit or add one which is missing for some reason, and I think we'd want that to propagate to this app, unless I've missed something.

lib/local_authorities_url_importer.rb
```diff
@@ -50,23 +49,4 @@ def parse_url(url)
       "http://" + url
     end
   end
-
```
jennyd: Could you rebase these removals back into the URL importer commit, so that they disappear from history?

lib/local_authorities_slug_importer.rb
```diff
@@ -0,0 +1,44 @@
...
+  include JsonDownloader
+
+  AUTHORITIES_URL = "https://raw.githubusercontent.com/alphagov/frontend/master/lib/data/authorities.json"
+
+  def self.import_slugs
+    new.import
+  end
+
+  def import
+    LocalAuthority.find_each do |authority|
```
jennyd: [`find_each` defaults to batches of 1000](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each), so unless we override that we aren't getting the benefit of using it here.

spec/lib/local_authorities_importer_spec.rb
sihugh: I'd figured that they wouldn't change unless it was in line with the GSS code.  But then they do don't they.  It's a fair point.

lib/local_authorities_importer.rb
sihugh: Is it worth updating Publisher to follow this guideline?  Possibly not if it's only for the short term... https://github.com/alphagov/publisher/blob/30b44b7417285589484d7e0d18c76329afd2920e/lib/local_interaction_importer.rb#L78

config/initializers/mapit.rb
sihugh: Again, we're going to have more connections, so keeping this tidy is a good idea.  Will tweak

lib/local-links-manager/import/local_authorities_importer.rb
```diff
@@ -38,6 +38,7 @@ def authorities_from_mapit
             Rails.logger.info("Updating authority #{mapit_la[:name]} (gss #{mapit_la[:gss]})")
             la.update_attributes!(
               name: mapit_la[:name],
+              snac: mapit_la[:snac],
```
jennyd: Now that we're updating the snac too, we're setting the same attributes on update and create. We could use `first_or_initialize.tap` to unify these cases so that the same code is used for both which will help us avoid accidentally handling them differently in future:

``` ruby
LocalAuthority.where(gss: mapit_la[:gss]).first_or_initialize.tap do |la|
  verb = la.persisted? ? "Updating" : "Creating"
  Rails.logger.info("#{verb} authority #{mapit_la[:name]} (gss #{mapit_la[:gss]})")

  la.name = mapit_la[:name]
  la.snac = mapit_la[:snac]
  la.tier = mapit_la[:tier]
  la.save!
end
```

The `tap` is there so that the block is run in both cases - `first_or_initialize` only uses the block when initializing and not when it finds an existing record.

Transition [uses this approach](https://github.com/alphagov/transition/blob/master/lib/transition/import/site_yaml_file.rb#L72-L90) in a very similar context - I know not everyone is fond of `tap`, though, so other opinions are available...

h-lame: We can avoid `.tap` entirely because we don't actually use the `LocalAuthority` that we build/update for anything. Something like:

``` ruby
la = LocalAuthority.where(gss: mapit_la[:gss]).first_or_initialize
verb = la.persisted? ? "Updating" : "Creating"
Rails.logger.info("#{verb} authority #{mapit_la[:name]} (gss #{mapit_la[:gss]})")

la.name = mapit_la[:name]
la.snac = mapit_la[:snac]
la.tier = mapit_la[:tier]
la.save!
```

There's something to be said for the block giving us some "encapsulation" around that work though, at the minor cost of understanding what `.tap` is doing (and why we're not using the block form of `first_or_initialize`).

lib/local_authorities_importer.rb
jennyd: I don't think it is worth it there - we'll be taking that import out of Publisher in the next few months, and that's the only place where it calls MapIt directly (it goes via Imminence to get the [areas](https://github.com/alphagov/publisher/blob/30b44b7417285589484d7e0d18c76329afd2920e/app/models/area.rb) for business support schemes).

lib/local-links-manager/import/local_authorities_importer.rb
sihugh: I might go halfway and not tap, but extract into a create_or_update method.

lib/local-links-manager/import/local_authorities_url_importer.rb
```diff
@@ -0,0 +1,56 @@
+require 'csv'
+
+module LocalLinksManager
+  module Import
+    class LocalAuthoritiesUrlImporter
```
issyl0: Is it worth adding "URL" to the inflections file, as it is generally understood to be ALL CAPS? Only the first letter in capitals looked weird to me when I first saw this class name, and we have prior art for it in [Transition](https://github.com/alphagov/transition/blob/master/config/initializers/inflections.rb#L23).

spec/rails_helper.rb
sihugh: I've moved this into support/shoulda.rb :+1:

lib/local_authorities_importer.rb
sihugh: I've used a hash instead.  Thanks :-)

lib/local_authorities_importer.rb
sihugh: Good plan, thanks. I've used a hash instead

lib/local_authorities_importer.rb
sihugh: Added CTY, thanks

lib/local_authorities_importer.rb
sihugh: I remapped this to use a hash so we can keep just the one set of area types as you suggested.  COI is listed as unitary.

lib/local_authorities_importer.rb
sihugh: I'm now using `gds-api-adapters` after @jennyd 's suggestion so we get around that.

lib/local_authorities_slug_importer.rb
sihugh: I took the second of these options.  It's not a vast json file, but it works nicely.  Thanks

lib/local_authorities_url_importer.rb
sihugh: I changed this importer to work less like the csv importers in Publisher (where it came from) and more like the other ones in this app that work on json.  We're no longer creating temp files which tidies this up.

app/views/layouts/application.html.erb
sihugh: I've removed it now anyway as this is now the front page

lib/tasks/local_authorities.rake
sihugh: Have nested these inside `local-links-manager/import` as per Transition's layout

lib/tasks/local_authorities.rake
sihugh: I've renamed these to `import_authorities`, `add_slugs` and `add_urls` which implies dependencies better

lib/local_authorities_url_importer.rb
sihugh: Yep, thanks

lib/local_authorities_slug_importer.rb
sihugh: I was prematurely optimising. :-/  Thanks

h-lame:
:+1: for the Scotocentric nature of the fixture data.

app/models/local_authority.rb
```diff
@@ -0,0 +1,5 @@
+class LocalAuthority < ActiveRecord::Base
+  validates :gss, :snac, uniqueness: true
+  validates :gss, :name, :snac, :tier, presence: true
+  validates :homepage_url, format: { with: URI.regexp }, allow_blank: true
+end
```
jennyd, line 7: We should also validate that `tier` is one of `%w{district county unitary}`, for completeness.

lib/local-links-manager/import/local_authorities_url_importer.rb
sihugh: @jennyd , @h-lame and I a chat about this last night.  I think we're going to change the class name to fit with "URL", but not change inflections (at least for the moment).  If we have a pressing need to do so later on, we can address it then.

app/models/local_authority.rb
sihugh, line 7: Seems fair, will add.

jennyd:
🔒 ⏰ 🍆 💭 🍚 👔 ✨

PR #6, opened 2016-04-27
Create models for importing services & interactions

emmabeynon:
- Created models for services, interactions and mapped service and interactions.
- This is in preparation for importing LGSL and LGIL lists.
- [Trello Card](https://trello.com/c/r8yv6gjk/368-import-all-services-and-interactions-into-local-links-manager-3)


db/migrate/20160427110629_create_service_interactions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateServiceInteractions < ActiveRecord::Migration
+  def change
+    create_table :service_interactions do |t|
+      t.integer :service_identifier
+      t.string :service_label
+      t.integer :ineraction_identifier
+      t.string :interaction_label
```
h-lame: It would probably be more useful if instead of storing the service and interaction identifiers and labels we stored the database ids of the Service and Interaction instances that have those identifiers.  This would let us hook up Service and Interaction via rails associations.

It would make importing more complex as we'd have to make sure we'd imported all the services and interactions first before importing the allowed combinations.  It does however give us scope for errors where try to import a combination of service 123 and interaction 456 but one (or both) of them don't exist yet.  This'll help us expose holes in the data (which _should_ be none as the service, interaction, and valid combinations lists all come from the same source, but you never know).

db/migrate/20160427110629_create_service_interactions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateServiceInteractions < ActiveRecord::Migration
+  def change
+    create_table :service_interactions do |t|
+      t.integer :service_identifier
+      t.string :service_label
+      t.integer :ineraction_identifier
```
h-lame: That said, if we do keep the service + interaction details in here, there's a typo here; it should be `interaction_identifier` but we've dropped a `t`.

spec/factories/services.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :service do
+    identifier ""
+    label "MyString"
+  end
```
h-lame, line 5: For all of the factories it would be nice if there were real looking examples for each field instead of the blank string or "MyString" defaults.

app/models/service_interaction.rb
```diff
@@ -0,0 +1,2 @@
+class ServiceInteraction < ActiveRecord::Base
+end
```
h-lame, line 7: Even if we don't add the associations it'd be worth adding simple validations for what is and isn't allowed.  For example: bare minimum we probably don't want to save a `Service` without an identifier and label, but maybe there's a specific format that the label should be in, or that the identifier should be in a specific range of numbers (I'm pretty sure we'd want the presence validations, less sure about the format + range ones - but we should investigate a bit).  There's probably similar validations we could add to `Interaction` and `ServiceInteraction`.

spec/models/service_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Service, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
h-lame: If we do add validations or associations then we can put something in these specs to cover them.  Consider using the [`shoulda-matchers` gem](https://github.com/thoughtbot/shoulda-matchers) for testing that sort of thing.  If you've never used it there are a bunch of other gds apps that use it so we can crib from them, but the docs are pretty good starting point too.

h-lame:
I know this is an "unblocking" PR so it doesn't need to be fully-featured or anything, but I can imagine that both the import and display stories would be made easier with things like associations and validations in place.

app/models/service_interaction.rb
jennyd, line 7: The identifiers and labels for both services and interactions should be unique, so we should validate them for both presence and uniqueness at least 🎁 ❄️

app/models/service_interaction.rb
```diff
@@ -0,0 +1,2 @@
+class ServiceInteraction < ActiveRecord::Base
```
jennyd, line 1: Does anyone have any suggestions for a better name for this model? Links will be related to it so it's probably something we'll be referring to frequently.

`Transaction` might work, but I think it's likely to be confusing when we start matching these up to published local transactions (and not all interactions are transactional anyway). Also actual database transactions...

We can rename it later so don't need to find the perfect name right away, but it's worth thinking about now.

emmabeynon:
@jennyd @h-lame We've made the necessary changes based on your comments - let us know if there's anything else.

app/models/service_interaction.rb
```diff
@@ -1,2 +1,4 @@
 class ServiceInteraction < ActiveRecord::Base
+  validates :service_identifier, :service_label, :interaction_identifier,
+  :interaction_label, presence: true, uniqueness: true
```
h-lame: This validation says that `service_identifier`, `service_label`, `interaction_identifier`, and `interaction_label` should all be present and should all be unique.  The first part is correct, but what we really want with uniqueness checking is not that the individual parts are unique, but that the combination is unique.  E.g. we should allow a mapping between one service (pay your council tax) and multiple interactions (find out about, pay, appeal) and we should allow the same interaction (find out about) to be mapped to multiple services (pay your council tax, dog wardens, school places).

Happily, rails gives us a `scope` option on the uniqueness validation to define saw that an attribute should be unique in combination with another, so we'll want to use that.  The [rails guides for the uniqueness validation](http://guides.rubyonrails.org/active_record_validations.html#uniqueness) are a good starting point for learning more.

spec/rails_helper.rb
```diff
@@ -59,3 +59,10 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :rspec
+    with.library :rails
+  end
+end
```
h-lame: This is my fault somewhat, but [Simon's PR](https://github.com/alphagov/local-links-manager/pull/4/files) adds shoulda-matchers already so we'll have a conflict to resolve when his is merged and we re-work this one on top of it.  If we're lucky git will resolve it, but we should check that we haven't duplicated this config, particularly as Simon's version lives in `spec/support/shoulda.rb` instead of `spec/rails_helper.rb` (it's better to keep `spec/spec_helper.rb` and `spec/rails_helper.rb` as pristine as possible to make it easier to incorporate changes to these files when we upgrade rspec and rspec-rails).

There's nothing we need to do now just something to be aware of when we rework this once Simon's is merged.

db/migrate/20160427110426_create_services.rb
```diff
@@ -0,0 +1,10 @@
+class CreateServices < ActiveRecord::Migration
+  def change
+    create_table :services do |t|
+      t.integer :identifier
```
jennyd: The ESD datasets call this number the identifier, but we've been referring to it between ourselves as the LGSL code (and LGIL code for the interaction one). It would be good to use the name that our users call it - maybe Jane can help us with that question.

app/models/service_interaction.rb
```diff
@@ -1,4 +1,12 @@
...
   validates :service_identifier, :service_label, :interaction_identifier,
   :interaction_label, presence: true, uniqueness: true
+
+  def self.find_service(identifier)
+    Service.find_by identifier: identifier
+  end
+
+  def self.find_interaction(identifier)
+    Interaction.find_by identifier: identifier
+  end
```
h-lame: Probably my fault for not being clearer when I suggested it, but what I expected here is that we would have methods on the instance rather than the class.  E.g. we could do:

``` ruby
service_interaction = ServiceInteraction.find(...)

service_interaction.service # returns the Service instance with the same identifier as service_instance.service_identifier

service_interaction.interaction # returns the Interaction instance with the same identifier as service_instance.interaction_identifier
```

This gives us an API similar to what we'd get if we used ActiveRecord belongs_to / has_one style associations (and leaves the door open to us doing that in the future if we think it makes sense).  The implementation would be almost identical, and the tests too, just not defined as `def self.find_blah` and not taking an argument.

spec/models/service_interaction_spec.rb
```diff
@@ -1,12 +1,33 @@
...
+    it { is_expected.to validate_uniqueness_of(:interaction_identifier) }
+    it { is_expected.to validate_uniqueness_of(:interaction_label) }
+  end
+
+  describe '#find_service' do
+    it 'returns the correct service from the Services database' do
+      service = FactoryGirl.create(:service, label: 'Abandoned shopping trolleys', identifier: 1152)
+      service_interaction = FactoryGirl.create(:service_interaction)
+
+      expect(ServiceInteraction.find_service(service_interaction.service_identifier)).to eq(service)
```
h-lame: It would be nice to be more explicit in the specs to make it clear _why_ we expect the method to return that particular service instance.  As it stands we could make this spec pass if the implementation was just to return `Service.first`.  For example we could be more explicit when building the `service` and `service_integration` instances so that they share identifiers, we could also add a negative test to show that it won't just return any `Service` instance, but only the one that matches on the identifier.

In general it's worth being explicit in our specs about what we're testing, that goes for the setup of the objects we use, but also for the name of the spec too.  Here we're just saying "returns the correct service" but we haven't clearly stated what "correct" means.  It makes it easier to come back to things later and work out what we meant - I'll often look at the specs for an object to discover what it's supposed to do before I look at the object implementation itself as specs can be clearer to read because they tend to be simpler (set up some objects, do a thing with one or more of them, check the results) compared to the actual implementations which can be super complex.

This comment goes for the `find_interaction` spec too.

app/models/interaction.rb
```diff
@@ -0,0 +1,3 @@
+class Interaction < ActiveRecord::Base
+  validates :identifier, :label, presence: true, uniqueness: true
+end
```
h-lame, line 5: Now that we've got uniqueness and presence validations it's worth backing those up with not-null and uniqueness constraints at the database level in our migrations.  [This thoughtbot article](https://robots.thoughtbot.com/the-perils-of-uniqueness-validations) is a pretty good discussion about why you might want to add db constraints as well as doing application-level validation.

db/migrate/20160427110629_create_service_interactions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateServiceInteractions < ActiveRecord::Migration
+  def change
+    create_table :service_interactions do |t|
+      t.integer :service_identifier
+      t.string :service_label
```
davidbasalla: Do we need to add `service_label` and `interaction_label` here? I would assume that these would always be the label of the referred service/interaction? (knowing full well what they say about 'assume' :) ). So just wondering if we can just keep the identifiers and get rid of the labels... 

Edit: Keeping the labels appears to complicate the uniqueness validations as well. Presumably we'll want to allow ServiceInteractions with same `service_identifier` and `service_label`. We only care about the uniqueness of the `service_identifier` and `interaction_identifier` combination...?

emmabeynon: @jennyd Any thoughts on this?

app/models/service_interaction.rb
davidbasalla: @h-lame Would there be any drawbacks to using `belongs_to` / `has_one` for this now?

db/migrate/20160427110629_create_service_interactions.rb
jennyd: I would expect that we could have only 2 columns on this table, `service_id` and `interaction_id`, referring to the `id` columns of services and interactions, like foreign keys. I don't think we should duplicate the `label` or `identifier` fields in this table - it would be more work to keep them up to date when they change in future.

app/models/service_interaction.rb
sihugh, line 1: ServiceInstance?  (Suggested, but it's a bit generic :-/ )

h-lame:
Other than the name for `ServiceInteraction` this looks ready to merge.  Are we happy (resigned to might be a better way of phrasing it?) with the name?  For now at least; it's not impossible to rename later if we have a fit of inspiration.

sihugh:
I can't think of any other names which aren't more horrid or potentially confusing.  I vote for merging.

PR #7, opened 2016-04-27
Add a startup.sh file for consistency with other apps

issyl0:
- This allows it to be run easily without bowler.


h-lame:
consistency FTW :+1:

PR #8, opened 2016-04-28
Add empty application.js

jennyd:
This file was being requested but returning 404.


h-lame:
Wonder how that happened?  We should have one with all the default rails comments at the top.

PR #9, opened 2016-05-06
Import services, interactions and service interactions

davidbasalla:
This PR adds importer classes and rake tasks to import Services, Interactions and ServiceInteractions from standards.esd website. The original files are in CSV format, so we download them with Net:HTTP's `get_response` method, parse them and store them as records in our local database. The order of the imports is important, as Services and Interactions have to be imported first, before ServiceInteractions (which acts as a join model between the two) can be imported successfully.

We split out a `CsvDownloader` class which we use in our importer classes as well as `LocalAuthorityUrlImporter`.

For ServiceInteractions we log a summary of imported rows, as there are a number of errors that can appear.

Worked with @klssmith and @emmabeynon on this story, just plastered my face on all commits as we were using my laptop :)

### DIFF
--- a/lib/local-links-manager/import/service_interactions_importer.rb
+++ b/lib/local-links-manager/import/service_interactions_importer.rb
@@ -13,16 +13,25 @@ module LocalLinksManager
 
       def initialize(csv_downloader = CsvDownloader.new(CSV_URL))
         @csv_downloader = csv_downloader
+        @missing_record_count = 0
+        @missing_id_count = 0
+        @created_or_updated_record_count = 0
       end
 
       def import_records
-        @csv_downloader.download.each do |row|
+        downloaded_csv_rows.each do |row|
           begin
             create_or_update_record(find_associated_records(parsed_hash(row)))
-          rescue MissingRecordError, MissingIdentifierError => e
+            @created_or_updated_record_count += 1
+          rescue MissingRecordError => e
+            @missing_record_count += 1
+            Rails.logger.error e.message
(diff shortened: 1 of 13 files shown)


h-lame:
This all looks good to me.  Although the description of the pr still talks about `open_uri` whereas the implementation now uses `Net::HTTP`.

lib/local-links-manager/import/service_interactions_importer.rb
```diff
@@ -70,6 +81,14 @@ def create_or_update_record(parsed_hash)
 
         service_interaction.save!
       end
+
+      def import_summary
+        "ServiceInteraction Import complete\n"\
+        "Downloaded CSV rows: #{downloaded_csv_rows.count}\n"\
+        "Imported records: #{ServiceInteraction.count}\n"\
```
h-lame: This isn't the number of imported records, it's the number of service interactions in the system.  I'm splitting hairs a bit, but we could have 10 rows in the CSV and 100 rows already imported, which is going to make this number look weird.  What we probably want here is `downloaded_csv_rows.count - missing_id_count - missing_record_count` - or we keep track of the successful rows like we do for the missing id and missing record ones.

davidbasalla: Thank you, have addressed your comment. I went with another count variable for successful rows added, mainly because I'm not 100% sure that the code catches all errors (there might be more than missing ids and missing associated records)

issyl0:
Seems fine. ⭐

PR #10, opened 2016-05-11
UI for services and interactions

issyl0:
- This means routes of the form
  `/local_authorities/aberdeen/services/16-to-19-bursary-fund/interactions`,
  using slugs as IDs to appear more nicely to users, and hence with
  nesting of services under local authorities, and interactions under
  services. To do the friendly URLs, we had to add a `slug` column to
  the Services table and generate the slug in the import script using
  `label.parameterize`.
- The interactions list only shows the applicable ones for the selected
  service. These aren't links yet - that's a separate story.
- Also a separate story is scoping the services list based on the ones
  that the Local Authority provides - now it just shows all 900.

Screenshots of now (2016-05-13):

screenshot 2016-05-13 16 26 08

screenshot 2016-05-13 16 26 20

screenshot 2016-05-13 16 26 35

### DIFF
--- a/app/controllers/interactions_controller.rb
+++ b/app/controllers/interactions_controller.rb
@@ -1,0 +1,6 @@
+class InteractionsController < ApplicationController
+  def index
+    @service = Service.find_by_slug!(params[:service_slug])
+    @interactions = @service.interactions
+  end
+end
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -1,0 +1,6 @@
+class ServicesController < ApplicationController
+  def index
+    @authority = LocalAuthority.find_by_slug!(params[:local_authority_slug])
+    @services = Service.all
+  end
+end
--- a/app/views/local_authorities/index.html.erb
+++ b/app/views/local_authorities/index.html.erb
@@ -13,7 +13,13 @@
       locals: {placeholder: "Filter list of local authorities"} %>
     </thead>
     <tbody>
-      <%= render @authorities %>
+      <% @authorities.each do |authority| %>
+        <tr>
+          <td class="name">
+            <%= link_to authority.name, local_authority_services_path(authority.slug), class: 'js-open-on-submit' %>
+          </td>
+        </tr>
+      <% end %>
     </tbody>
   </table>
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160511144329) do
+ActiveRecord::Schema.define(version: 20160513113110) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -54,6 +54,7 @@ ActiveRecord::Schema.define(version: 20160511144329) do
     t.string   "label",      null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string   "slug",       null: false
     t.string   "tier"
   end
 
--- a/app/models/interaction.rb
+++ b/app/models/interaction.rb
@@ -2,4 +2,5 @@ class Interaction < ActiveRecord::Base
   validates :lgil_code, :label, presence: true, uniqueness: true
 
   has_many :service_interactions
+  has_many :services, through: :service_interactions
 end
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -1,8 +1,9 @@
 class Service < ActiveRecord::Base
-  validates :lgsl_code, :label, presence: true, uniqueness: true
+  validates :lgsl_code, :label, :slug, presence: true, uniqueness: true
   validates :tier, inclusion: { in: %w{all county/unitary district/unitary}, allow_nil: true }
 
   has_many :service_interactions
+  has_many :interactions, through: :service_interactions
 
   scope :for_tier, ->(tier) {
     case tier
--- a/app/views/interactions/index.html.erb
+++ b/app/views/interactions/index.html.erb
@@ -1,0 +1,32 @@
+<% content_for :page_title, @service.label %>
+<div class="page-title">
+  <h1><%= @service.label %></h1>
+</div>
+
+<% if @interactions.any? %>
+  <table class="table table-striped table-hover table-bordered contacts-table" data-module="filterable-table">
+    <thead>
+      <tr class="table-header">
+        <th>LGIL Code</th>
+        <th>Local Government Interactions (<%= @interactions.count %>)</th>
+      </tr>
+      <%= render partial: "govuk_admin_template/table_filter",
+      locals: {placeholder: "Filter list of local interactions"} %>
+    </thead>
+    <tbody>
+      <% @interactions.each do |interaction| %>
+        <tr>
+          <td class="lgil_code">
+            <%= interaction.lgil_code %>
+          </td>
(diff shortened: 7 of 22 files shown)


h-lame:
For the general approach: :+1: - I think we'll want these nested views, but I also think there's an argument for keeping the services root too, so maybe we want both.

app/controllers/local_authorities_controller.rb
```diff
@@ -2,4 +2,8 @@ class LocalAuthoritiesController < ApplicationController
   def index
     @authorities = LocalAuthority.all.order(name: :asc)
   end
+
+  def show
+    @authority = LocalAuthority.find_by_slug(params[:slug])
```
h-lame: `find_by_slug` will return `nil` if the provided slug doesn't map to an actual authority so we should prefer using the bang! variant `find_by_slug!` which will raise an `ActiveRecord::RecordNotFound` exception.  By default rails will catch those exceptions and render a 404, whereas returning a `nil` will most likely end up causing a `NoMethodError` exception somewhere and rendering 500.

app/views/local_authorities/index.html.erb
```diff
@@ -13,7 +13,13 @@
       locals: {placeholder: "Filter list of local authorities"} %>
     </thead>
     <tbody>
-      <%= render @authorities %>
+      <% @authorities.each do |authority| %>
+        <tr>
+          <td class="name">
+            <%= link_to authority.name, local_authority_path(authority.slug), class: 'js-open-on-submit' %>
```
h-lame: What does this class do?

app/views/local_authorities/index.html.erb
```diff
@@ -13,7 +13,13 @@
       locals: {placeholder: "Filter list of local authorities"} %>
     </thead>
     <tbody>
-      <%= render @authorities %>
+      <% @authorities.each do |authority| %>
```
h-lame, line 16: The previous version of this delegated to another view for rendering each authority object (the view in question is `app/views/local_authorities/_local_authority.html.erb` - the [rails api documentation for partial rendering](http://api.rubyonrails.org/classes/ActionView/PartialRenderer.html) and in particular the bit about ["the default case"](http://api.rubyonrails.org/classes/ActionView/PartialRenderer.html#class-ActionView::PartialRenderer-label-Rendering+the+default+case) has more details). 

I don't have strong opinions on which is better; inlining the loop (what we have now) or using another partial (what we had before), but we should clear up the old partial if we prefer inlining, or go back to using the partial if we prefer using partials.

app/controllers/services_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ServicesController < ApplicationController
   def show
     @service = Service.find(params[:id])
+    @interactions = Interaction.joins(:service_interactions).where('service_interactions.service_id': @service.id)
```
h-lame: It feels like this is the sort of association walking we'll want to do a bunch, so it'll be useful to add these as activerecord associations to the `Service` and `Interaction` models to make it easier.

``` ruby
class Service
  has_many :service_interactions
  has_many :interactions, through: :service_interactions
end
```

And the same for `Interaction` to get all the `Service`s that implement it.

app/controllers/interactions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class InteractionsController < ApplicationController
+  def index
+    # We parameterized this label to make it look like a slug for
+    # the URL, so now we need to turn it back into a string so that
+    # the service can be found from it. There's probably a better way.
+    @service = Service.find_by_label(params[:service_label].downcase.gsub("-"," "))
```
h-lame: You're right that there's probably a better way, it's a bit more work though: add a slug field to services and generate it on create. It's unlikely that this reverse parameterization is going to do the right thing, parameterisation destroys information (e.g. `"Pavements - personal injury".parameterize` gives `"pavements-personal-injury"` but reversing it `"pavements-personal-injury".downcase.gsub("-"," ")` gives `"pavements personal injury"`)

(resurrecting a comment left on a non-PR version of this that commit that has since been rebased away - apols if you got this twice)

app/views/local_authorities/index.html.erb
issyl0: This class lives in the admin template and is so that when pressing enter after typing in the filter box, as some people have a habit of doing, the first link in the filtered list will open without having to explicitly click it. If you're wondering how it works in practice, check out the organisations list in Transition.

app/views/local_authorities/index.html.erb
issyl0, line 16: Ah. I saw that and wondered what it was for! I'll remove the partial.

app/views/local_authorities/index.html.erb
h-lame: Ah!  I did a search on github but could only find it in views; forgot to look in govuk admin template for the js implementation :(.  Thanks!

app/controllers/interactions_controller.rb
issyl0: OK. Adding a slug field wouldn't be too bad, and would make URLs look nicer for users, so I'll work on that.

fofr:
I think the codes should appear to the left of their titles, then they can be read together easily.

fofr:
The nested views should have a breadcrumb to allow easy navigation back up the chain.

app/controllers/interactions_controller.rb
issyl0: Done!

issyl0:
Thanks @fofr. I've reordered the columns. I'll do breadcrumbs in a separate PR.

app/controllers/interactions_controller.rb
```diff
@@ -0,0 +1,6 @@
+class InteractionsController < ApplicationController
+  def index
+    @service = Service.find_by_slug!(params[:service_slug])
+    @interactions = Service.find(@service.id).interactions
```
h-lame: This can be `@interactions = @service.interactions` - we've already got the `Service` instance (we even use it in this 2nd `Service.find`) so there's no need to look it up twice.

issyl0: Thanks! 😴

db/schema.rb
```diff
@@ -54,6 +54,7 @@
     t.string   "label",      null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string   "slug"
```
h-lame: Given we're using this field for lookups it will be useful if we have a unique index on it to make that faster.  It might also be good to have it as a non-null column while we're at it.  We can back those up with the relevant validations and specs too.  Doing all that means we'll want to add some extra code to the migration as well to make sure any existing rows won't be invalid.

Something like:

``` ruby
add_column :services, :slug, unique: true

Service.all.each do |service|
  service.slug = service.label.parameterize
  service.save!
end

change_column_null :services, :slug, false
```

Although it's worth reading this [blog post about using models in migrations](https://coderwall.com/p/zav1dg/using-models-in-rails-migrations) too, as there are reasons not to just use `Service` directly (tl;dr so that the migrations can run in a future world where `Service` has been renamed or removed).

issyl0: I thought about that after I tried to add a not null constraint and it failed because I hadn't rerun the import yet, but thought that could be added and the services migrated at a later date. May as well do it now though now you've mentioned it. 😉

issyl0: Done! Thanks!

h-lame:
Looks great.

brenetic:
🏆

PR #12, opened 2016-05-11
Use CSV header converters

davidbasalla:
In the Importer classes (Service, Interaction, ServiceInteraction), we previously converted some header names to more readable names (such as "Mapped identifier" -> :lgil_code). This forced another level of indirection for imported CSV rows, having to call the `parsed_hash` method on a row first before it could get passed to the `create_or_update_method`.

This commit makes use of the `CSV` library's support for [converters](http://ruby-doc.org/stdlib-2.0.0/libdoc/csv/rdoc/CSV.html#method-i-header_convert), so we can pass a hash of conversions to the `CsvDownloader` class, and the conversions are applied as the CSV rows are imported. This helps to declutter the Importer classes, makes the `CsvDownloader` (rightfully) responsible for the conversions and makes the tests more readable.

### DIFF
--- a/lib/local-links-manager/import/csv_downloader.rb
+++ b/lib/local-links-manager/import/csv_downloader.rb
@@ -5,12 +5,16 @@ class CsvDownloader
   class DownloadError < Error; end
   class MalformedCSVError < Error; end
 
-  def initialize(csv_url)
+  def initialize(csv_url, header_conversions = {})
     @csv_url = csv_url
+    @header_conversions = header_conversions
   end
 
   def download
-    CSV.parse(downloaded_csv, headers: true)
+    CSV.parse(downloaded_csv,
+              headers: true,
+              header_converters: field_name_converter)
+
   rescue CSV::MalformedCSVError => e
     raise MalformedCSVError, "Error #{e.class} parsing CSV in #{self.class}"
   end
@@ -25,5 +29,11 @@ private
     end
 
     response.body
+  end
+
+  def field_name_converter
+    lambda do |field|
+      @header_conversions.key?(field) ? @header_conversions[field] : field
+    end
   end
 end
--- a/lib/local-links-manager/import/interactions_importer.rb
+++ b/lib/local-links-manager/import/interactions_importer.rb
@@ -4,17 +4,21 @@ module LocalLinksManager
   module Import
     class InteractionsImporter
       CSV_URL = "http://standards.esd.org.uk/csv?uri=list/interactions"
+      FIELD_NAME_CONVERSIONS = {
+        "Label" => :label,
+        "Identifier" => :lgil_code
+      }
 
       def self.import
         new.import_records
       end
 
-      def initialize(csv_downloader = CsvDownloader.new(CSV_URL))
+      def initialize(csv_downloader = CsvDownloader.new(CSV_URL, FIELD_NAME_CONVERSIONS))
         @csv_downloader = csv_downloader
       end
 
       def import_records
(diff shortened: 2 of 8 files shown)


h-lame:
I like this.  Getting rid of the parsed_hash method feels like a big improvement.

lib/local-links-manager/import/csv_downloader.rb
```diff
@@ -26,4 +33,10 @@ def downloaded_csv
 
     response.body
   end
+
+  def name_converter(conversions)
```
h-lame: I don't know that `name_converter` is the most intention revealing name.  Not that I think I have a better one...  Not too keen on overusing `header` either, so it's not like `header_converters` would be better.

lib/local-links-manager/import/csv_downloader.rb
```diff
@@ -26,4 +33,10 @@ def downloaded_csv
...
+    lambda do |field|
+      conversions.detect { |k, _| k == field } ? conversions[field] : field
```
h-lame: We can simplify this to `conversions.fetch(field, field)` which I admit is a little opaque.  Maybe `conversions.key?(field) ? conversions[field] : field` would be less opaque while still a bit simpler than using `detect`.

lib/local-links-manager/import/csv_downloader.rb
```diff
@@ -5,12 +5,19 @@ class Error < RuntimeError; end
   class DownloadError < Error; end
   class MalformedCSVError < Error; end
 
-  def initialize(csv_url)
+  def initialize(csv_url, header_conversions = {})
     @csv_url = csv_url
+
+    if header_conversions
+      CSV::HeaderConverters[:name_converter] = name_converter(header_conversions)
```
h-lame: Could we avoid setting a named converter entirely and use the lambda directly:

``` ruby
def initialize(csv_url, header_conversions = {})
  @csv_url = csv_url
  @header_conversions = header_conversions
end

def download
  CSV.parse(downloaded_csv, headers: true, header_converters: name_converter)
end

def name_converter
  lambda do |field|
    @header_conversions.detect { |k, _| k == field } ? @header_conversions[field] : field
  end
end
```

By setting it as a named converter in `CSV::HeaderConverters[:name_converter]` we expose that conversion globally.  Crucially, if one of our `CsvDownloader` instances does use header conversions (and so sets the `:name_converter` here) but the instance of `CsvDownloader` doesn't use header conversions (and so doesn't set the `:name_converter` here) then this 2nd instance would end up using the `:name_converter` set by the previous instance.  It'll get even more weird if we had multiple `CsvDownloader` instances running at the same time in threads - ultimately the last `:name_converter` set would be used by all instances.

lib/local-links-manager/import/csv_downloader.rb
davidbasalla: yeah I struggled with finding a good name. Went with `field_name_converter` in the end, as that seems to match CSV terminology somewhat (without using the word 'header' again)?

lib/local-links-manager/import/csv_downloader.rb
davidbasalla: Great, thanks!

lib/local-links-manager/import/csv_downloader.rb
davidbasalla: Yeah, that's definitely better! thanks!

davidbasalla:
Thank you @h-lame for the comments, have addressed and rebased. I've also made the 'field_name_conversion` hash into a constant for each Importer class.

h-lame:
Changes look good to me.  Merging.

PR #13, opened 2016-05-12
Add tiers to services

h-lame:
For: https://trello.com/c/duUE8jPS/375-add-support-for-la-tiers-to-local-links-manager-3

`Service` now has a `tier` field that contains `"all"`, `"district/unitary"`, `"county/unitary"` or is `nil`.  We import values for this from the [`local_services.csv` that lives in publisher](https://github.com/alphagov/publisher/blob/master/data/local_services.csv).  As mentioned in the commits it seems very likely that this file originally came from https://data.gov.uk/dataset/local-directgov-url-instructions and has since been updated.  What's not hugely clear where the tier information in that dataset comes from.

### DIFF
--- a/lib/local-links-manager/import/services_tier_importer.rb
+++ b/lib/local-links-manager/import/services_tier_importer.rb
@@ -1,0 +1,90 @@
+require_relative 'csv_downloader'
+
+module LocalLinksManager
+  module Import
+    class ServicesTierImporter
+      CSV_URL = "https://raw.githubusercontent.com/alphagov/publisher/master/data/local_services.csv"
+      FIELD_NAME_CONVERSIONS = {
+        'LGSL' => :lgsl_code,
+        'Providing Tier' => :tier,
+      }
+      class MissingRecordError < RuntimeError; end
+      class MissingIdentifierError < RuntimeError; end
+
+      def self.import
+        new.import_tiers
+      end
+
+      def initialize(csv_downloader = CsvDownloader.new(CSV_URL, FIELD_NAME_CONVERSIONS))
+        @csv_downloader = csv_downloader
+        @csv_rows = 0
+        @missing_record_count = 0
+        @missing_id_count = 0
+        @invalid_record_count = 0
+        @updated_record_count = 0
+        @ignored_rows_count = 0
+      end
+
+      def import_tiers
+        with_each_csv_row do |row|
+          counting_errors do
+            if update_record(row)
+              @updated_record_count += 1
+            else
+              @ignored_rows_count += 1
+            end
+          end
+        end
+        Rails.logger.info import_summary
+      end
+
+    private
+
+      def import_summary
+        "ServicesTier Import complete\n"\
+        "Downloaded CSV rows: #{@csv_rows}\n"\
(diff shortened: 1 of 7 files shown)


spec/models/service_spec.rb
```diff
@@ -13,4 +13,62 @@
...
+
+  describe '#provided_by?' do
+    let(:district) { FactoryGirl.build(:local_authority, tier: 'district') }
+    let(:county) { FactoryGirl.build(:local_authority, tier: 'county') }
+    let(:unitary) { FactoryGirl.build(:local_authority, tier: 'unitary') }
+
+    context 'when the tier is blank' do
+      subject { FactoryGirl.build(:service, tier: nil) }
+      it 'is false for a district council' do
+        expect(subject).not_to be_provided_by(district)
```
davidbasalla, line 54: I hadn't noticed this before... is `be_provided_by` a wrapper for `provided_by?(authority)`? Does Rspec generate this automatically?

h-lame, line 54: Yeah, for any `predicate?` method on an instance rspec lets you use `be_predicate` as the matcher: https://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/built-in-matchers/predicate-matchers

lib/local-links-manager/import/services_tier_importer.rb
```diff
@@ -0,0 +1,93 @@
...
+        @missing_id_count = 0
+        @invalid_record_count = 0
+        @updated_record_count = 0
+        @ignored_rows_count = 0
+      end
+
+      def import_tiers
+        with_each_csv_row do |row|
+          counting_errors do
+            if update_record(parsed_hash(row))
```
davidbasalla: This could be rewritten now since https://github.com/alphagov/local-links-manager/pull/12 got merged, making the `CsvDownloader` class do the field_name_conversions, thereby not having to call the `parsed_hash` method on `row` here

h-lame: Yup.  On it as we speak.

lib/local-links-manager/import/services_tier_importer.rb
```diff
@@ -0,0 +1,93 @@
...
+          @csv_rows += 1
+          block.call(row)
+        end
+      rescue CsvDownloader::Error => e
+        Rails.logger.error e.message
+      rescue => e
+        Rails.logger.error "Error #{e.class} importing in #{self.class}\n#{e.backtrace.join("\n")}"
+      end
+
+      def counting_errors(&block)
```
davidbasalla, line 76: Nice! I like how this splits out the error counting from the update loop

lib/local-links-manager/import/services_tier_importer.rb
h-lame: Rebased this away.

davidbasalla:
Looks good! 👍 I ran the import script locally too and all seems to work nicely!

issyl0:
I was a bit surprised that this doesn't print anything to the console when it ends, but this behaviour is consistent across all of our rake tasks. It worked though! 👍

PR #14, opened 2016-05-16
Restrict services for authority ui to those provided by tier

h-lame:
For: https://trello.com/c/duUE8jPS/375-add-support-for-la-tiers-to-local-links-manager-3

When clicking through to a Local Authority show page we limit to show only those services that the LA should provide according to our tier data (see: #13).

We only have tier data for ~250 out of ~1000 services so there's now no way of seeing all those "un-tiered" services in the system UI.

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -4,4 +4,8 @@ class LocalAuthority < ActiveRecord::Base
   validates :homepage_url, format: { with: URI.regexp }, allow_blank: true
   validates :tier, inclusion: { in: %w(county district unitary),
     message: "%{value} is not an allowed tier" }
+
+  def provided_services
+    Service.for_tier(self.tier)
+  end
 end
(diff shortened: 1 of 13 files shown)


app/models/local_authority.rb
```diff
@@ -4,4 +4,8 @@ class LocalAuthority < ActiveRecord::Base
   validates :homepage_url, format: { with: URI.regexp }, allow_blank: true
   validates :tier, inclusion: { in: %w(county district unitary),
     message: "%{value} is not an allowed tier" }
+
+  def provided_services
+    Service.for_tier(self.tier)
+  end
```
issyl0, line 10: Oh, that's how the limiting to local authority works: by tiers. I was wondering how we'd do that.

spec/factories/local_authorities.rb
```diff
@@ -4,7 +4,7 @@
     gss "S12000041"
     snac "00QC"
     tier "unitary"
-    slug "angus"
+    slug { name.parameterize }
```
issyl0, line 7: Nice!

PR #15, opened 2016-05-19
Import and display links

h-lame:
For: https://trello.com/c/Q0fMMggq/370-import-local-links-into-local-links-manager-5

In which we create a `Link` class to store the url that a given local authority uses for a given service interaction.  We display these links on the local authority service interactions index page as follows:

aberdeen-homelessness-support

They are imported from the same CSV that publisher uses.  Note that we're not currently: 
1. ignoring 'x' urls (although we do ignore those rows without matching snacs, lgsl, or lgil codes, so this might cover all of those - certainly no 'x' urls appear to have been imported locally).
2. doing anything to handle deletion (links removed in local direct gov admin no longer appear in the csv - see [Publisher's ghost detector](https://github.com/alphagov/publisher/blob/master/lib/local_authority_interaction_ghost_detector.rb))

Best reviewed commit-by-commit mostly because the last commit is very noisy and distorts the "full diff"

### DIFF
--- a/lib/local-links-manager/import/csv_downloader.rb
+++ b/lib/local-links-manager/import/csv_downloader.rb
@@ -5,30 +5,50 @@ class CsvDownloader
   class DownloadError < Error; end
   class MalformedCSVError < Error; end
 
-  def initialize(csv_url, header_conversions = {})
+  def initialize(csv_url, header_conversions: {}, encoding: 'windows-1252')
     @csv_url = csv_url
     @header_conversions = header_conversions
+    @encoding = encoding
   end
 
-  def download
-    CSV.parse(downloaded_csv,
-              headers: true,
-              header_converters: field_name_converter)
+  def each_row(&block)
+    download do |csv|
+      csv.each(&block)
+    end
+  end
 
+  def download(&block)
+    downloaded_csv do |data|
+      block.call(
+        CSV.parse(
+          data,
+          headers: true,
+          header_converters: field_name_converter
+        )
+      )
+    end
   rescue CSV::MalformedCSVError => e
     raise MalformedCSVError, "Error #{e.class} parsing CSV in #{self.class}"
   end
 
 private
 
-  def downloaded_csv
-    response = Net::HTTP.get_response(URI.parse(@csv_url))
+  def downloaded_csv(&block)
+    Tempfile.create(['local_links_manager_import', @csv_url.gsub(/[^0-9A-z.\-]+/, '_'), 'csv']) do |temp_file|
+      temp_file.set_encoding('ascii-8bit')
 
-    unless response.code_type == Net::HTTPOK
-      raise DownloadError, "Error downloading CSV in #{self.class}"
+      response = Net::HTTP.get_response(URI.parse(@csv_url))
+
+      unless response.code_type == Net::HTTPOK
+        raise DownloadError, "Error downloading CSV in #{self.class}"
+      end
+
+      temp_file.write(response.body)
+
+      temp_file.rewind
+      temp_file.set_encoding(@encoding, 'UTF-8')
+      block.call temp_file
     end
-
-    response.body
   end
 
   def field_name_converter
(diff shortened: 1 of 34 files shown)


spec/support/table_row_matcher.rb
```diff
@@ -0,0 +1,51 @@
+module TableRowMatchers
+  class SimpleTableRowMatcher < Capybara::RSpecMatchers::Matcher
```
davidbasalla, line 2: This is really cool! Is this a standard thing to do and are there any docs for how to do this (couldn't find much on a quick google)?

h-lame, line 2: http://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/custom-matchers is a good starting point.  The reason this doesn't use the short-hand style

``` ruby
RSpec::Matchers.define :be_a_multiple_of do |expected|
  match do |actual|
    actual % expected == 0
  end
end
```

is that I wanted it to work with capybara page objects and that requires a bit more effort.  Happily there's a Capybara rspec matcher base class to use and we can take "inspiration" from the existing matchers that Capybara provides (see: https://github.com/jnicklas/capybara/blob/master/lib/capybara/rspec/matchers.rb).

In general defining your own matchers for complex expectations can be quite powerful as it makes the specs read much nicer.  I'm often guilty of forgetting that this is possible or desirable though.  Only remembered this time because it was so verbose without writing a custom matcher.

spec/models/service_interaction_spec.rb
```diff
@@ -13,4 +13,57 @@
...
+      expect(described_class.find_by_lgsl_and_lgil(100, 1)).to be_nil
+    end
+
+    it 'returns nil if the supplied lgil is not a valid interaction' do
+      FactoryGirl.create(:service, lgsl_code: 100, label: 'Service 100')
+
+      expect(described_class.find_by_lgsl_and_lgil(100, 1)).to be_nil
+    end
+
+    context 'to avoid n+1 queries' do
```
davidbasalla, line 51: Nice! I don't see tests that check for this too often ⭐

brenetic:
Can we pull the last commit out into a separate PR?

h-lame:
@brenetic not really - the links file won't import without the encoding handling fixes, and the solution for the encoding handling will leave files lying around if we don't use the block form :(

lib/local-links-manager/import/csv_downloader.rb
```diff
@@ -5,30 +5,50 @@ class Error < RuntimeError; end
...
     raise MalformedCSVError, "Error #{e.class} parsing CSV in #{self.class}"
   end
 
 private
 
-  def downloaded_csv
-    response = Net::HTTP.get_response(URI.parse(@csv_url))
+  def downloaded_csv(&block)
+    Tempfile.create(['local_links_manager_import', @csv_url.gsub(/[^0-9A-z.\-]+/, '_'), 'csv']) do |temp_file|
+      temp_file.set_encoding('ascii-8bit')
```
davidbasalla, line 38: I know this is coming from Publisher, but could you elaborate why we're setting this particular encoding here? (desperately need to catch up on my 'encoding' reading :/ )

davidbasalla:
This all looks good to me! Quite a bit of stuff to review, so might have missed some subtleties, but 👍 from me

lib/local-links-manager/import/csv_downloader.rb
h-lame, line 38: My guess would be that because there are no invalid byte sequences in ascii-8bit - so we can write any bytes we'd like without getting an encoding error.  Then we can later when we're ready to read the data out (L49 below) we set it to convert from the encoding we think it is into utf8 as we read.  TBH: I'm not sure why we don't set it to the encoding we think it is straight off - but I assume there's some valid reason.  I'll see if the git logs for publisher shed any light on this approach.

davidbasalla:
Needs a rebase, happy to merge once done

h-lame:
Rebased for your merging pleasure.

PR #16, opened 2016-05-20
Breadcrumbs

h-lame:
Add breadcrumbs to the app (like what @fofr asked for on #10).

breadcrumbs

We use the [gretel gem](https://github.com/lassebunk/gretel) following the lead of [transition](https://github.com/alphagov/transition), but instead of [manually rendering the crumbs](https://github.com/alphagov/transition/blob/master/app/views/layouts/application.html.erb#L76-L82), we configure gretel to render `bootstrap` styled crumbs and it all comes out ok.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,9 @@ gem 'capistrano-rails', group: :development
 
 gem 'pg'
 
+# Provides breadcrumbs, see config/breadcrumbs.rb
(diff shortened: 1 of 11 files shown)


davidbasalla:
🆒!

sihugh:
Loving the gif!

PR #18, opened 2016-05-25
Display enabled services for each authority

klssmith:
- Added rake task to get the enabled services csv from Publisher and enable any Service currently imported.
- Log any enabled services that are in the csv but not currently imported.
- Changed the default encoding for csv importing to UTF-8.
- Change local authority services view to only show enabled services.

[Trello card](https://trello.com/c/LhMS1J3H/388-mark-services-to-be-enabled-for-use-on-gov-uk)

Mobbed by @emmabeynon @klssmith @brenetic @h-lame

### DIFF
--- a/db/migrate/20160523155343_add_enabled_column_to_service.rb
+++ b/db/migrate/20160523155343_add_enabled_column_to_service.rb
@@ -1,0 +1,5 @@
+class AddEnabledColumnToService < ActiveRecord::Migration
+  def change
+    add_column :services, :enabled, :boolean, default: false, null: false
+  end
+end
--- a/lib/local-links-manager/import/csv_downloader.rb
+++ b/lib/local-links-manager/import/csv_downloader.rb
@@ -5,7 +5,7 @@ class CsvDownloader
   class DownloadError < Error; end
   class MalformedCSVError < Error; end
 
-  def initialize(csv_url, header_conversions: {}, encoding: 'windows-1252')
+  def initialize(csv_url, header_conversions: {}, encoding: 'UTF-8')
     @csv_url = csv_url
     @header_conversions = header_conversions
     @encoding = encoding
--- a/lib/local-links-manager/import/enabled_service_checker.rb
+++ b/lib/local-links-manager/import/enabled_service_checker.rb
@@ -1,0 +1,40 @@
+require_relative 'csv_downloader'
+
+module LocalLinksManager
+  module Import
+    class EnabledServiceChecker
+      CSV_URL = "https://raw.githubusercontent.com/alphagov/publisher/master/data/local_services.csv"
+
+      def self.enable
+        new.enable_services
+      end
+
+      def initialize(csv_downloader = CsvDownloader.new(CSV_URL))
+        @csv_downloader = csv_downloader
+      end
+
+      def enable_services
+        @csv_downloader.each_row do |row|
+          set_enabled_service(row)
+        end
+      rescue CsvDownloader::Error => e
+        Rails.logger.error e.message
+      rescue => e
+        Rails.logger.error "Error #{e.class} enabling in #{self.class}\n#{e.backtrace.join("\n")}"
+      end
+
+    private
+
+      def set_enabled_service(row)
+        service = Service.find_by(lgsl_code: row["LGSL"])
+        if service.nil?
+          Rails.logger.warn("'#{row['LGSL']}' is not an imported Service")
+        else
+          service.enabled = true
+          Rails.logger.info("'#{row['LGSL']}' enabled")
+          service.save!
+        end
+      end
+    end
+  end
+end
--- a/lib/tasks/import/service_interactions.rake
+++ b/lib/tasks/import/service_interactions.rake
@@ -2,6 +2,7 @@ require 'local-links-manager/import/services_importer'
 require 'local-links-manager/import/interactions_importer'
 require 'local-links-manager/import/service_interactions_importer'
 require 'local-links-manager/import/services_tier_importer'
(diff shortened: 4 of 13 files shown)


lib/local-links-manager/import/csv_downloader.rb
```diff
@@ -5,7 +5,7 @@ class Error < RuntimeError; end
   class DownloadError < Error; end
   class MalformedCSVError < Error; end
 
-  def initialize(csv_url, header_conversions: {}, encoding: 'windows-1252')
+  def initialize(csv_url, header_conversions: {}, encoding: 'UTF-8')
```
davidbasalla, line 8: I think this makes the `LocalAuthoritiesURLImporter` csv import fail. You may have to update [this code](https://github.com/alphagov/local-links-manager/blob/master/lib/local-links-manager/import/local_authorities_url_importer.rb#L12) to be the following:

```
def initialize(csv_downloader = CsvDownloader.new(CSV_URL, encoding: 'windows-1252'))
```

lib/local-links-manager/import/enabled_service_checker.rb
```diff
@@ -0,0 +1,38 @@
...
+        end
+      rescue CsvDownloader::Error => e
+        Rails.logger.error e.message
+      rescue => e
+        Rails.logger.error "Error #{e.class} enabling in #{self.class}\n#{e.backtrace.join("\n")}"
+      end
+
+    private
+
+      def check_enabled_services(row)
```
davidbasalla: Quite possibly nitpicking, but I'm wondering whether this method could be renamed to `enable_service`? Singular as opposed to plural because we're only dealing with one service at a time. And removing the 'check' part of the name as it does not express the intent entirely.

lib/local-links-manager/import/enabled_service_checker.rb
```diff
@@ -0,0 +1,38 @@
+module LocalLinksManager
```
davidbasalla, line 3: I think this is missing `require_relative 'csv_downloader'`.

spec/lib/local-links-manager/import/enabled_service_checker_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    context 'when the csv is downloaded successfully' do
+      before do
+        stub_csv_rows(csv_rows)
+      end
+
+      it 'sets enabled to true for required services' do
+        LocalLinksManager::Import::EnabledServiceChecker.new(csv_downloader).enable_services
+        service = Service.find_by(lgsl_code: 1614)
+        expect(service.enabled).to eq(true)
```
davidbasalla: Could this be 

```
expect(service_0.reload.enabled).to eq(true)
expect(service_1.reload.enabled).to eq(true)
```

instead?

davidbasalla:
Looks good 👍 , only some minor comments. 

Have you considered making this rake task be part of the 'import_all' task for ServiceInteractions?

lib/local-links-manager/import/csv_downloader.rb
klssmith, line 8: Thanks - we've changed this

lib/local-links-manager/import/enabled_service_checker.rb
klssmith: We've changed the method name to `set_enabled_service` now, thanks

klssmith:
@davidbasalla We've made all the changes now, so if you could have another look that would be great

lib/tasks/import/service_interactions.rake
```diff
@@ -32,5 +34,10 @@ namespace :import do
     task add_service_tiers: :environment do
       LocalLinksManager::Import::ServicesTierImporter.import
     end
+
+    desc "Enable services used on Gov.uk"
+    task enable_all: :environment do
```
davidbasalla: Could we rename the task to something like `enable_services`? Right now because of the namespacing it sounds like we're enabling all ServiceInteractions instead of Services.

db/migrate/20160523155343_add_enabled_column_to_service.rb
```diff
@@ -0,0 +1,5 @@
+class AddEnabledColumnToService < ActiveRecord::Migration
+  def change
+    add_column :services, :enabled, :boolean, default: false, null: false
```
davidbasalla, line 3: More of an open question - would you typically add an index to this kind of column? As it will be used fairly often...?

h-lame, line 3: A good question.  Part of me thinks: "yeah put an index on it because we'll be filtering by this a lot", but part of me thinks "no need, it's a boolean field so there's only two values, surely an index won't help" - so I'm not sure.  Did some digging (on stackoverflow, obvs) and found this http://stackoverflow.com/questions/12539772/adding-an-index-on-a-boolean-field which suggests a partial index might be useful?  Never heard of partial indexes before so that's exciting.

In conclusion: maybe?

davidbasalla, line 3: So I tried adding an index to the 'enabled' column, but running `EXPLAIN ANALYZE` in dev seems to indicate that the query plan does not make use of the index. Presumably because over 25% of all services are enabled, so too many rows are returned and a sequential scan proves to be more performant.

I also tried changing the existing indexes for the Service table to be partial indexes, but did not see any improvement in query times. I'm aware that the query planner might behave differently in the production environment, but I'm happy to just merge this PR as is (ie not adding any extra index), and revisit this sort of optimisation in future when required.

PR #19, opened 2016-05-31
Enable editing of links and homepage URLs

brenetic:
- Added ability to edit homepage URL's
- Enable the editing of links
# Screenshots

screen shot 2016-06-03 at 10 57 59

screen shot 2016-06-03 at 11 01 01

screen shot 2016-06-03 at 10 55 45

screen shot 2016-06-03 at 10 55 32

screen shot 2016-06-03 at 10 56 00

[Trello editing links](https://trello.com/c/qtpaquan/389-enable-editing-of-links)
[Trello editing homepage url](https://trello.com/c/DfDlIIRc/390-make-the-homepage-url-for-each-local-authority-editable)

Mobbed by: @emmabeynon @sihugh @klssmith @brenetic @issyl0

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,31 +1,25 @@
 class LinksController < ApplicationController
+  before_action :load_dependencies
+
   def edit
-    load_dependencies
-    @link = Link.get_link(params)
+    @link = Link.retrieve(params)
   end
 
   def update
-    @link = Link.get_link(params)
+    @link = Link.retrieve(params)
     @link.url = params[:link][:url]
     validate_and_save(@link)
   end
 
   def create
-    @link = Link.new
-    @link.local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
-    @link.service_interaction = ServiceInteraction.find_by(
-      service: Service.find_by(slug: params[:service_slug]),
-      interaction: Interaction.find_by(slug: params[:interaction_slug])
-    )
+    @link = Link.build(params)
     @link.url = params[:link][:url]
-
     validate_and_save(@link)
   end
 
 private
 
   def validate_and_save(link)
-    load_dependencies
     if link.save
       flash[:success] = "Link has been saved."
       flash[:lgil] = @interaction.lgil_code
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -5,10 +5,6 @@ class LinksController < ApplicationController
     @interaction = Interaction.find_by(slug: params[:interaction_slug])
 
     @link = Link.get_link(params)
-    if @link.nil?
-      @link = Link.new
-      @link.url = 'n/a'
-    end
   end
 
   def update
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,39 +1,43 @@
 class LinksController < ApplicationController
   def edit
-    @local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
-    @service = Service.find_by(slug: params[:service_slug])
-    @interaction = Interaction.find_by(slug: params[:interaction_slug])
-
+    load_dependencies
     @link = Link.get_link(params)
   end
 
   def update
-    link = Link.get_link(params)
-    link.url = params[:link][:url]
-    validate_and_save(link)
+    @link = Link.get_link(params)
+    @link.url = params[:link][:url]
+    validate_and_save(@link)
   end
 
   def create
-    link = Link.new
-    link.local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
-    link.service = Service.find_by(slug: params[:service_slug])
-    link.interaction = Interaction.find_by(slug: params[:interaction_slug])
-    link.service_interaction = ServiceInteraction.find_by(service: link.service, interaction: link.interaction)
-    link.url = params[:link][:url]
+    @link = Link.new
+    @link.local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+    @link.service_interaction = ServiceInteraction.find_by(
+      service: Service.find_by(slug: params[:service_slug]),
+      interaction: Interaction.find_by(slug: params[:interaction_slug])
+    )
+    @link.url = params[:link][:url]
 
-    validate_and_save(link)
+    validate_and_save(@link)
   end
 
 private
 
   def validate_and_save(link)
-    if link.valid?
-      link.save!
+    if link.save
       flash[:success] = "Link has been saved."
       redirect_to local_authority_service_interactions_path(service_slug: params[:service_slug])
     else
-      flash[:danger] = "Please enter a valid link."
-      redirect_to action: "edit"
+      load_dependencies
+      flash.now[:danger] = "Please enter a valid link."
+      render :edit
     end
+  end
+
+  def load_dependencies
+    @local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+    @interaction = Interaction.find_by(slug: params[:interaction_slug])
+    @service = Service.find_by(slug: params[:service_slug])
   end
 end
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,5 +1,9 @@
 class LinksController < ApplicationController
   def edit
+    @local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+    @service = Service.find_by(slug: params[:service_slug])
+    @interaction = Interaction.find_by(slug: params[:interaction_slug])
+
     @link = Link.get_link(params)
     if @link.nil?
       @link = Link.new
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,10 +1,39 @@
 class LinksController < ApplicationController
-
   def edit
     @link = Link.get_link(params)
     if @link.nil?
       @link = Link.new
       @link.url = 'n/a'
+    end
+  end
+
+  def update
+    link = Link.get_link(params)
+    link.url = params[:link][:url]
+    validate_and_save(link)
+  end
+
+  def create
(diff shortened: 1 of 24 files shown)


app/controllers/links_controller.rb
```diff
@@ -0,0 +1,43 @@
+class LinksController < ApplicationController
+  def edit
+    @local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+    @service = Service.find_by(slug: params[:service_slug])
+    @interaction = Interaction.find_by(slug: params[:interaction_slug])
+
+    @link = Link.get_link(params)
+    if @link.nil?
+      @link = Link.new
+      @link.url = 'n/a'
```
issyl0: This could be shortened to `@link = Link.new(url: 'n/a')`.

app/controllers/links_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    link.interaction = Interaction.find_by(slug: params[:interaction_slug])
+    link.service_interaction = ServiceInteraction.find_by(service: link.service, interaction: link.interaction)
+    link.url = params[:link][:url]
+
+    validate_and_save(link)
+  end
+
+private
+
+  def validate_and_save(link)
```
issyl0: ActiveRecord checks validations by default on `save` - http://guides.rubyonrails.org/active_record_validations.html#when-does-validation-happen-questionmark - do we need this method?

spec/controllers/links_controller_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe LinksController, type: :controller do
```
davidbasalla, line 3: Should we add some tests for the controller actions here?

app/controllers/links_controller.rb
davidbasalla: I'm wondering whether we should leave `@link.url` as `nil` here and move the logic for displaying 'n/a' to the view or a helper like LinksHelper? To me showing 'n/a' is more a question of what we display when we get a `nil` link. Changing the value of `@link.url` to 'n/a' technically means that it's not `nil` anymore. Not sure if I'm making sense. What do you think?

davidbasalla: Also I haven't tried it yet but could we use a kind of `first_or_create`/`find_or_create_by` pattern here?

issyl0: @davidbasalla Now you mention that, yep!

sihugh: I think we were happy with not persisting links that don't have a real URL in them, but it sounds good.  I think moving the `if nil?` use `n/a` to a helper is a good idea.

app/controllers/links_controller.rb
sihugh: The main bit of this wasn't firing the validations so much as the redirections and flash messages.  I expect these will change a bit after the show and tell.  We could compress the `if link.valid?` and `link.save!` lines into just `if link.save` though.

spec/controllers/links_controller_spec.rb
sihugh, line 3: We probably should.  We have got feature tests that cover this though 👍

app/controllers/links_controller.rb
davidbasalla: One more question on this, do we even need to set this to be 'n/a'? Would it be okay just to leave `@link.url` as `nil`? The field in the form would then just be empty which still indicates that there is no url for this link.

sihugh: Good question!  We should ask the rest of the hive brain 🐝

app/views/links/edit.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <h2><%= @interaction.label %></h2>
+</div>
+<b>LGSL</b> <%= @service.lgsl_code %>
+<hr>
+<b>LGIL</b> <%= @interaction.lgil_code %>
+<hr>
+
+<%= label_tag 'link', 'URL' %>
+
+<%= form_for @link, url: {action: "update"} do |f| %>
```
h-lame: I don't think this `url: {action: "update"}` is doing anything as when `@link` is unsaved we end up in the `create` action, and when it is saved we end up in the `update` action.  This is the default behaviour of `form_for` so our `url` param isn't doing anything (and acts to confuse the reader as they'll think we always end up in the `update` action which isn't the case).

That said, if our intention is to always go through the `update` action regardless of it being a saved or unsaved (not a terrible idea if we can do something with @davidbasalla's idea of using some kind of `first_or_intialize` in / around our `Link.get_link` call) then this is the right sort of thing to do, but it's not working - maybe worth exploring this a little bit more.

app/controllers/links_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    if link.valid?
+      link.save!
+      flash[:success] = "Link has been saved."
+      redirect_to local_authority_service_interactions_path(service_slug: params[:service_slug])
+    else
+      flash[:danger] = "Please enter a valid link."
+      redirect_to action: "edit"
```
h-lame: Because we redirect here we lose a load of information about what the user has done.
1. We lose the url that they entered - so when we render the form again we can only show them the url from the db (or "n/a")
2. We lose the `link` instance that we called `.valid?` on which means we lose the error messages on there - when rails renders a form using `form_for` the form helpers will look at the object we passed to that to decide if the input elements should be given error classes to style them differently.

This is because a redirect causes a whole new http request/response to happen and the `link` instance and user entered params only exist for the request/response they were created in.  If instead of redirecting we asked to render the edit form again (`render :edit`) we'd keep the `link` instance so that rendering the form could use the validation errors and the url that the user had entered already.  Note that the form wants to use `@link` but our `create` action only uses `link`...

app/controllers/links_controller.rb
h-lame: +1 for keeping the 'n/a' out of the controller

Further to @davidbasalla's comment about `first_or_create` it might be interesting to make `Link.get_link` _always_ return a `Link` instance.  In the case where the provided arguments allow us to fetch something from the DB the behaviour remains the same, but when those args don't resolved to a row in the DB we could build an unsaved `Link` instance and set it up with appropriate `LocalAuthority` and `ServiceInteraction` associations (e.g. what we do in the `create` action) so that we _could_ save it once we've filled in the url.

I don't know that this is a great idea, but I think it's interesting, and it's a pattern I've used before.

app/controllers/links_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    link.url = params[:link][:url]
+    validate_and_save(link)
+  end
+
+  def create
+    link = Link.new
+    link.local_authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+    link.service = Service.find_by(slug: params[:service_slug])
+    link.interaction = Interaction.find_by(slug: params[:interaction_slug])
+    link.service_interaction = ServiceInteraction.find_by(service: link.service, interaction: link.interaction)
```
h-lame: We don't need to set `service`, `interaction`, and `service_interaction`.  If we set `service_interaction` then we'll get `service` and `interaction` for "free" because our `service` and `interaction` associations come _through_ the `service_interaction` association.

app/controllers/local_authorities_controller.rb
```diff
@@ -2,4 +2,27 @@ class LocalAuthoritiesController < ApplicationController
...
+
+  def update
+    authority = LocalAuthority.find_by(slug: params[:slug])
+    authority.homepage_url = params[:local_authority][:homepage_url]
+    validate_and_save(authority)
+  end
+
+private
+
+  def validate_and_save(authority)
```
h-lame, line 18: What's been said already about `valid?` and `save!` and `redirect_to` in the `LinksController` applies here too.

app/models/link.rb
```diff
@@ -14,4 +14,12 @@ class Link < ActiveRecord::Base
...
       .where(service_interactions: { service_id: service })
   }
+
+  def self.get_link(params)
+    self.joins(:local_authority, :service, :interaction).find_by(
+      local_authorities: { slug: params[:local_authority_slug] },
+      services: { slug: params[:service_slug] },
+      interactions: { slug: params[:interaction_slug] }
+    )
+  end
```
h-lame, line 34: @brenetic said that when doing this as a scope it would return loads of `Link` instances instead of just one.  The reason is that scopes _always_ return an array (techincally it's an `ActiveRecord::Relation` instance - which is what allows us to chain them together) - this is even if you've used a method like `.first` or `.find_by` which usually returns a single instance.  When you write it as a class method then there's none of the "magic" scope stuff happening to make sure we always return something chainable so in this case we would always get a single instance (or nil if one can't be found).

app/controllers/links_controller.rb
sihugh: Thanks for the comments. We have addressed this now. Will update in a bit

spec/controllers/links_controller_spec.rb
h-lame, line 3: FWIW: controller tests are going away in rails 5 (http://blog.bigbinary.com/2016/04/19/changes-to-test-controllers-in-rails-5.html explains the changes in detail - but not the rationale.  rspec is going to do the work that means you'll still be able to write them with rspec so when/if we upgrade we should be able to keep them, but the writing is on the wall that directly testing your controller is a no-longer-the-rails-way kinda thing to do so it might get harder and harder to do.

This is not to say that we shouldn't test our controllers, but that the specifics of how rails told you to test your controllers in the past (checking which template was rendered, checking which variables are assigned to share with the view, etc..) are not the way to do it.  Instead rails 5 suggests you test controllers with integration tests (feature specs in rspec parlance) where you don't reach inside to check the internals, but you can test the "full-stack" by checking what's on the rendered views.

So, all this to say, we have feature specs already so unless there's some complex branching route through the controller that we haven't covered in those, it's probably better to not introduce controller specs.

davidbasalla, line 3: Thanks for sharing, good to know!

PR #20, opened 2016-05-31
Add govuk-lint to Rakefile

davidbasalla:
Inspired by similar setup in specialist-publisher-rebuild, this seems like a useful thing to do as it allows us to catch linting errors earlier


brenetic:
😱 👮

PR #22, opened 2016-06-07
Delete links

brenetic:
- Now able to delete all links
- Only show the Delete button if the link exists
- Show 'Edit link' button if link exists otherwise 'Add link'
- Added .no-gutter css to remove padding for editing form page
# Before

screen shot 2016-06-07 at 12 32 16
# After

screen shot 2016-06-06 at 15 50 04

screen shot 2016-06-07 at 12 32 46

screen shot 2016-06-07 at 12 35 13

[Trello card](https://trello.com/c/XuQM5Zi7/400-enable-deleting-a-link)

Mobbed by @sihugh @brenetic @emmabeynon @klssmith @issyl0

### DIFF
--- a/app/helpers/links_helper.rb
+++ b/app/helpers/links_helper.rb
@@ -6,4 +6,12 @@ module LinksHelper
       'No link'
     end
   end
+
+  def interaction_button_text(interaction)
+    @links.key?(interaction) ? 'Edit link' : 'Add link'
+  end
+
+  def homepage_button_text(authority)
+    authority.homepage_url.empty? ? 'Add link' : 'Edit link'
+  end
 end
--- a/app/views/interactions/index.html.erb
+++ b/app/views/interactions/index.html.erb
@@ -7,6 +7,11 @@
 <div class="page-title">
   <h2><%= @service.label %></h2>
 </div>
+
+<% if flash[:success_action] %>
+<div class="alert alert-success"><%= flash[:success_action] %></div>
+<% end %>
+
 <p><b>LGSL</b> <%= @service.lgsl_code %></p>
 
 <% if @interactions.any? %>
--- a/app/views/interactions/index.html.erb
+++ b/app/views/interactions/index.html.erb
@@ -34,7 +34,7 @@
           <td>
           </td>
           <td class='text-center'>
-            <%= link_to('Edit link', edit_local_authority_service_interaction_links_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s") %>
+            <%= link_to(interaction_button_text(interaction), edit_local_authority_service_interaction_links_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s") %>
           </td>
         </tr>
       <% end %>
--- a/app/views/links/edit.html.erb
+++ b/app/views/links/edit.html.erb
@@ -6,22 +6,33 @@
   <h2><%= @interaction.label %></h2>
 </div>
 
-<% if flash[:bad_url] %>
-<div class="alert alert-danger">Please enter a valid link.</div>
+<% if flash[:failed_action] %>
+<div class="alert alert-danger"><%= flash[:failed_action] %></div>
 <% end %>
 
 <p><b>LGSL</b> <%= @service.lgsl_code %><b class="add-left-margin">LGIL</b> <%= @interaction.lgil_code %></p>
 
-<%= label_tag 'link', 'URL' %>
-
-<p class="text-muted">eg http://www.local-authority.gov.uk</p>
-
 <%= form_for @link, url: local_authority_service_interaction_links_path(@local_authority.slug, @service.slug, @interaction.slug), method: "put" do |f| %>
-  <div class="form-group <% if flash[:bad_url] %>has-error<% end %>">
-    <%= f.text_field :url, class: "form-control", value: @link.url%>
-    <div class='actions add-vertical-margins'>
+  <div class="form-group <% if flash[:failed_action] %>has-error<% end %>">
+    <div class='col-xs-12 no-gutter'>
+      <%= f.label :url, 'URL' %>
+
+      <p class="text-muted">eg http://www.local-authority.gov.uk</p>
+
+      <%= f.text_field :url, class: "form-control" %>
+    </div>
+
+    <div class='actions add-vertical-margins col-xs-6 no-gutter'>
       <%= link_to 'Cancel', local_authority_service_interactions_path(service_slug: @service.slug), class: 'btn btn-default add-right-margin' %>
       <button type='submit' class='btn btn-success'>Save</button>
     </div>
+  </div>
+<% end %>
+
+<% if @link.url %>
+  <div class='actions add-vertical-margins col-xs-6 no-gutter'>
+    <%= form_tag(local_authority_service_interaction_links_path(@local_authority.slug, @service.slug, @interaction.slug), method: :delete) do %>
+      <button name="submit" class="btn btn-danger pull-right">Delete</button>
+    <% end %>
   </div>
 <% end %>
--- a/app/views/links/edit.html.erb
+++ b/app/views/links/edit.html.erb
@@ -18,7 +18,7 @@
 
 <%= form_for @link, url: local_authority_service_interaction_links_path(@local_authority.slug, @service.slug, @interaction.slug), method: "put" do |f| %>
   <div class="form-group <% if flash[:bad_url] %>has-error<% end %>">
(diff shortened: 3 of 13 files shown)


app/views/links/edit.html.erb
```diff
@@ -18,7 +18,7 @@
 
 <%= form_for @link, url: local_authority_service_interaction_links_path(@local_authority.slug, @service.slug, @interaction.slug), method: "put" do |f| %>
   <div class="form-group <% if flash[:bad_url] %>has-error<% end %>">
-    <%= f.text_field :url, class: "form-control", value: @link.display_url%>
+    <%= f.text_field :url, class: "form-control", value: @link.url%>
```
h-lame: You probably don't need the `value: @link.url` bit anymore, as it would be the default because `@link` is the object we're using in the form and we're drawing an input field for its `url` attribute.

app/views/interactions/index.html.erb
```diff
@@ -29,12 +29,12 @@
...
           <td class="name">
             <%= interaction.label %>
-            <p><%= display_link_for(interaction) %></p>
+            <p><%= link = display_link_for(interaction) %></p>
           </td>
           <td>
           </td>
           <td class='text-center'>
-            <%= link_to('Edit link', edit_local_authority_service_interaction_links_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s") %>
+            <%= link_to(link == 'No link' ? 'Add link' : 'Edit link', edit_local_authority_service_interaction_links_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s") %>
```
h-lame: This is slightly unusual - we've coupled the text of the action button to the text we display for the link under the label.  If we change `display_link_for` to return some more markup, or just swap from `'No link'` to `'No Link'` we're going to have to change this conditional.  It'd be better to add a new helper that does something similar to `display_link_for` by looking for a `Link` instance for the supplied `interaction` and rendering the appropriate text based on finding or not finding that `Link`.

This helper would probably be v. similar to `display_link_for` and so could point us in the direction of wanting some kind of presenter for this page to wrap up the logic and get things both out of the view and out of helpers.  But maybe we don't need to yet - it's something we might consider later as this page gets more complex.

app/views/links/edit.html.erb
```diff
@@ -6,22 +6,33 @@
...
+<div class="alert alert-danger"><%= flash[:failed_action] %></div>
 <% end %>
 
 <p><b>LGSL</b> <%= @service.lgsl_code %><b class="add-left-margin">LGIL</b> <%= @interaction.lgil_code %></p>
 
-<%= label_tag 'link', 'URL' %>
+<%= form_for @link, url: local_authority_service_interaction_links_path(@local_authority.slug, @service.slug, @interaction.slug), method: "put" do |f| %>
+  <div class="form-group <% if flash[:failed_action] %>has-error<% end %>">
+    <div class='col-xs-12 no-gutter'>
+      <%= label_tag 'link', 'URL' %>
```
h-lame: The label we generate here looks like `<label for="link">URL</label>` which means it's the label for an input with the id `link`, but the input we generate has the id `link_url`.  If we replace this call with `<%= f.label :url, 'URL' %>` then it'll do the right thing.  In this version we ask the form builder to generate the label rather than rendering it ourselves (much like we swapped from `text_field_tag` to `f.text_field` to render the input field) and so it joins things up properly for us.  See: http://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-label for more info on the `f.label` helper specificially and that page for most of the form builder helpers you can call on `f` in general.

spec/controllers/links_controller_spec.rb
```diff
@@ -15,4 +15,15 @@
...
+
+  describe 'delete links' do
+    it 'handles deletion of links that have already been deleted' do
+      login_as_stub_user
+      delete :destroy, local_authority_slug: @local_authority.slug, service_slug: @service.slug, interaction_slug: @interaction.slug
+      expect(response).to have_http_status(302)
+      delete :destroy, local_authority_slug: @local_authority.slug, service_slug: @service.slug, interaction_slug: @interaction.slug
+      expect(response).to have_http_status(302)
+      expect(flash[:failed_action]).not_to be_present
+    end
```
h-lame, line 25: There's a weird quirk of controller tests/specs where trying multiple requests in a single test don't work properly.  But, I'm not sure we actually need the second request here?  Looking at the setup of the spec there isn't a `Link` instance yet, so the first request is already trying the deletion-of-a-link-that-doesn't-exist path, the second request seems redundant.

app/views/interactions/index.html.erb
```diff
@@ -7,6 +7,11 @@
 <div class="page-title">
   <h2><%= @service.label %></h2>
 </div>
+
+<% if flash[:success_action] %>
```
tijmenb, line 11: I think this duplicates the flash feature in `govuk_admin_template` (https://github.com/alphagov/govuk_admin_template#flash-messages). Or is that intentional?

spec/features/links/links_spec.rb
```diff
@@ -21,9 +21,9 @@
       expect(page).to have_table_row('4', 'Interaction 2 No link', '', 'Add link')
     end
 
-    it "shows 'No link' when editing a blank link" do
+    it "shows an empty cell when editing a blank link" do
       within('.table') { click_on('Add link', match: :first) }
-      expect(page).to have_field('link_url', with: 'No link')
+      expect(page.find_by_id('link_url').value).to be_blank
```
h-lame, line 26: It's always a shame to move away from the capybara helpers. Does `expect(page).to have_field('link_url', with: '')` not work?

app/views/interactions/index.html.erb
brenetic: We went down the route of a helper method similar to display_link_for but essentially felt that one call to a method doing a database lookup would be better than doing two were we to actually have a second helper method. Perhaps performance is not an issue?

app/views/interactions/index.html.erb
brenetic, line 11: Intentional. We just needed to change the position of it on the page

spec/features/links/links_spec.rb
brenetic, line 26: That was our first try, but unfortunately we could not get it to work

app/views/interactions/index.html.erb
h-lame: `display_link_for` doesn't talk to the database.  The controller fetches everything from the db and puts it in a hash (that's what `@links` is), and `display_link_for` looks up entries in that hash.

spec/features/links/links_spec.rb
h-lame, line 26: Boo! Ah well :(

app/views/interactions/index.html.erb
brenetic: That is a point so embarrassingly obvious that perhaps we should say no more :P

brenetic:
@h-lame  Thank you for the comments. Have addressed them all and updated

app/helpers/links_helper.rb
```diff
@@ -6,4 +6,8 @@ def display_link_for(interaction)
       'No link'
     end
   end
+
+  def button_text(interaction)
```
h-lame: Apols for massive nitpicking here, but `button_text` is a bit generic for the name here.  What button?  Every button? (I know, and you know, but maybe future us won't remember).  

We can be generic for methods on a class as the instance they live on will give us hints about the context they refer to (e.g. `Link.retreive` - this lives on `Link` so it's a strong sign it will `retreive` us some `Link` instances), but with helper methods we should be as explicit as possible because they all get mixed in together and made available everywhere in the view (even though `button_text` lives in the `LinksHelper` we don't ever call it via `LinksHelper` so we don't get that context).

h-lame:
We should have a think about flash messages - the names we use and the locations we display them.  We've got the ones that govuk_admin_template handles for us: (`success`, `info`, `warning`, and `danger`) and the ones we're using: (`success_action`, and `failed_action`).  The govuk_admin_template ones are rendered as part of the layout template, just before we render the content for each page, our ones are rendered somewhere inside our pages usually just below the `h1` element.  We're also not consistent: we use `success_action` for add/edit/delete `Link` isntances but `success` for editing the homepage url for `Service`, so the messages for doing similar things appear in different places.

We could ask govuk_admin_template not to render the flash messages and do something ourselves, perhaps by creating another named `content_for` block to render the `h1` and then always put the flash messages under that.  But that might be a bit unexpected.  Perhaps this needs some design thought.

Anyway, not going to hold this PR up because of this, just noting it down for now as a place to have a discussion.

PR #23, opened 2016-06-10
Change flash messages

brenetic:
- Set govuk_admin_template showing of flash messages to false in order to allow Local Links Manager to control where the flash message will appear.
- Remove custom flash messages and use the govuk_admin_template defaults.
- Strip leading and trailing whitespace when saving urls for both homepage urls and for service interaction links.

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -9,7 +9,7 @@ class LinksController < ApplicationController
     if @link.save
       redirect
     else
-      flash.now[:failed_action] = "Please enter a valid link."
+      flash.now[:danger] = "Please enter a valid link."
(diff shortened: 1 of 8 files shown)


jackscotti:
Would the use of default govuk_admin_template flash messages change the way flash messages look like?

brenetic:
Using the same classes for styling so should still look the same

sihugh:
@jackscotti : we're using the same markup as the template, we just want them closer to where the user has made the edit/action.  They're picking up the same styles, so it should be good!
(brenetic beat me to it)

jackscotti:
👍  thank you for the explanation

sihugh:
👍

PR #24, opened 2016-06-15
Delete links that are no longer in the CSV

sihugh:
When a local authority deletes a link from local.direct.gov.uk we need to
ensure that we delete it from our dataset as well.

We already did this for the imports in publisher, but we need to also do this
in local links manager until we stop importing data.  An alternative would be
to drop the data and recreate it each time, but this creates a period of
instability where there are no records.

The Publisher work is around this point:
https://github.com/alphagov/publisher/pull/436/commits/b025051977c402d3ede788720a0814f62890540c

We have a fail-safe total to ensure that we don't purge all our links should we
somehow end up with an empty CSV file.  We currently have 80_000 links in
Local Links Manager, however work continues to ascertain which services and
interactions we support, so I 'guessed' at a safe total of 40_000 to begin
with.

https://trello.com/c/KuN7jtEr/394-imports-should-delete-things-sensibly-5

### DIFF
--- a/lib/local-links-manager/import/links_importer.rb
+++ b/lib/local-links-manager/import/links_importer.rb
@@ -14,6 +14,9 @@ module LocalLinksManager
         95A 95B 95C 95D 95E 95F 95G 95H 95I 95J 95K 95L 95M
         95N 95O 95P 95Q 95R 95S 95T 95U 95V 95W 95X 95Y 95Z
       )
+      # We assume that if we can import more than this number of links then the
+      # CSV file was valid (i.e. not empty, truncated, or missing bits)
+      MIN_VALID_LINK_COUNT = 40_000
 
       class MissingRecordError < RuntimeError; end
       class MissingIdentifierError < RuntimeError; end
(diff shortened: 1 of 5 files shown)


spec/lib/local-links-manager/import/links_importer_spec.rb
```diff
@@ -104,6 +104,42 @@
...
+            {
+              lgil_code: '1',
+              lgsl_code: '123',
+              snac: '00AB',
+              url: 'http://www.example.com/first-link',
+            },
+          ]
+          stub_csv_rows(csv_rows_without_second_link)
+
+          ENV['MIN_LINK_COUNT'] = '1'
```
davidbasalla: Would it be worth adding another test here to document the behaviour when the `MIN_LINK_COUNT` is not met?

spec/lib/local-links-manager/import/links_importer_spec.rb
```diff
@@ -104,6 +104,42 @@
 
         expect(link.reload.url).to eq 'http://www.example.com/this-is-now-different'
       end
+
+      context 'when a link is missing from the CSV' do
+        it 'is removed from the database' do
+          service = FactoryGirl.create(:service, lgsl_code: 123, label: 'Service 123')
```
davidbasalla: The setup for creating a Link is quite involved. Wondering whether it would be worth it to make our factories do more of this work, so that we could just call something like this (without specifying the dependencies):

```
first_link = FactoryGirl.create(:link, url: 'http://example.com/first-link')
second_link = FactoryGirl.create(:link, url: 'http://example.com/second-link')
```

When creating more than one Link (as in this test), we'd have to set up [sequences](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#sequences) for our `LocalAuthority`, `Service` and `Interaction` factories due to their uniqueness validations on various fields. 

Anyone else have any opinions on this?

lib/local-links-manager/import/links_importer.rb
```diff
@@ -26,27 +26,59 @@ def initialize(csv_downloader = CsvDownloader.new(CSV_URL, header_conversions: F
         @missing_record_count = 0
         @missing_id_count = 0
         @invalid_record_count = 0
+        @deleted_record_count = 0
+        @links_in_csv = {}
```
h-lame: Minor: You could use a [`Set`](http://ruby-doc.org/stdlib-2.3.1/libdoc/set/rdoc/Set.html) here rather than using a hash like one.  I think you'll get the same performance characteristics, but the intent might be clearer - while reading the code I was momentarily confused about what the values in the hash were, when in fact we don't care about them at all, only the keys.

spec/lib/local-links-manager/import/links_importer_spec.rb
h-lame: It'd certainly help get some of the boilerplate out of the tests, so it sounds like a reasonable thing to me.

sihugh: Sounds like a plan, I'll put this in a separate commit and add to the PR

spec/lib/local-links-manager/import/links_importer_spec.rb
sihugh: :+1:

lib/local-links-manager/import/links_importer.rb
sihugh: Who knew that data structures could be helpful? 👍

sihugh:
Can we merge this?  I'd like to refactor the tests to extract some of the boilerplate, but haven't managed to get anything satisfactory yet.  I'm on 2nd line for a week now.

davidbasalla:
@sihugh What do you think about bf9007f042b2fe01fa8bd0149b82fdd3fcf8bae3 for reducing boilerplate in the tests?

PR #25, opened 2016-06-15
Import LocalAuthority slugs from mapit

davidbasalla:
As of alphagov/mapit#20 MapIt returns the code type 'govuk_slug' for local authorities. In this PR we update the `LocalAuthoritiesImporter` to set the slug for each local authority. It also means that we can remove the `LocalAuthoritiesSlugImporter` class.

For https://trello.com/c/nTjR73Ee/403-switch-local-links-manager-slug-import-to-use-slugs-from-mapit-2

Paired with @klssmith


issyl0:
Looks sane to me.

PR #26, opened 2016-06-16
Fix Northern Ireland blank homepage URL bug

klssmith:
- Fixed bug that prevented services page rendering for NI local authorities due to homepage URL being blank
- Fixed bug that displayed the current page path as the homepage URL on the services page instead of displaying no URL

### DIFF
--- a/app/helpers/links_helper.rb
+++ b/app/helpers/links_helper.rb
@@ -12,6 +12,6 @@ module LinksHelper
   end
 
   def homepage_button_text(authority)
-    authority.homepage_url.empty? ? 'Add link' : 'Edit link'
+    authority.homepage_url.blank? ? 'Add link' : 'Edit link'
   end
 end
--- a/app/views/shared/_local_authority_details.html.erb
+++ b/app/views/shared/_local_authority_details.html.erb
@@ -1,7 +1,7 @@
 <div class="page-title">
   <h1><%= authority.name %></h1>
   <p>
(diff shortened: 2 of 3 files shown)


jennyd:
Could you add tests for these bugfixes, so that we don't accidentally break them again in future? We need to be confident that we can render pages without raising errors for data which can be missing.

klssmith:
@jennyd There are now tests for both bugs, could you take a look to see if these are ok?

h-lame:
This fixes the issue so ➕ 💯 for that, but I wonder if we might want to make the UI more explicit when there is no homepage link.  When we have an LA with no homepage we leave the area blank, whereas for a service we explicitly say "No link" where the link would have gone (thanks to `LinksHelper#display_link_for` - https://github.com/alphagov/local-links-manager/blob/master/app/helpers/links_helper.rb#L2-L8).  We can't just reuse that method, but maybe doing something similar would be nice?

I'm not sure this comment is worth holding up the PR for though as fixing a thing is better than waiting to make it "perfect".  We might want some design input too.  Just a thought.

klssmith:
Yes, this pull request stops the wrong homepage link being displayed for local authorities where the homepage url is nil, but we still need to decide if something else should be displayed there instead.

h-lame:
Cool.  Let's not worry about it for this PR then.

davidbasalla:
Looks good to me. 👍

PR #27, opened 2016-06-20
Ignore known problems while importing links

h-lame:
For: https://trello.com/c/bXJG1RU8/401-investigate-and-fix-data-discrepancies-3

We know that the links csv from local direct gov has some rows that we
don't care about so we ignore those explicitly rather than mixing them
in with the unexpected errors.
- rows with a blank SNAC - these are rows for authorities that aren't
  local authorities, mostly they deal with issuing licenses, and we
  don't care about them
- rows with an 'x' url - these indicate that the authority provides the
  service but the url is special somehow.  It's mostly licenses and
  again we don't care
- rows for the old NI councils - these councils don't exist anymore
  and we don't care what their links were

In all cases we ignore the rows and increment the ignored_rows_count
variable rather than raising a more specific error.  To test this we
expose the counts as readers on the importer.

### DIFF
--- a/lib/local-links-manager/import/links_importer.rb
+++ b/lib/local-links-manager/import/links_importer.rb
@@ -10,6 +10,10 @@ module LocalLinksManager
         'LGIL' => :lgil_code,
         'Service URL' => :url
       }
+      OLD_NI_COUNCIL_SNACS = %w(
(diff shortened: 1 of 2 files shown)


brenetic:
👍

PR #28, opened 2016-06-30
Check link status code

davidbasalla:
We add a LinkChecker class that accepts an array of URL strings and returns an an array with the status and the time the link was checked. In the normal case, the status will be an HTTP status code like 200 or 404, but in various error cases, a custom error message will be returned such as 'Timeout Error' or 'SSL Error'. We use the [faraday gem](https://github.com/lostisland/faraday) to make the actual GET requests. We initially started with just the NetHttp library but found that redirects difficult to handle for all cases.

We add `LinkStatusUpdater` and `HomepageUpdater` classes which use the `LinkChecker` to make requests for the relevant URLs and write the results to the database.

For https://trello.com/c/KBT9RC9q/406-check-the-status-code-of-links

Mobbed with @klssmith and @emmabeynon

### DIFF
--- a/lib/local-links-manager/check_links/link_checker.rb
+++ b/lib/local-links-manager/check_links/link_checker.rb
@@ -1,0 +1,52 @@
+module LocalLinksManager
+  module CheckLinks
+    class LinkChecker
+      TIMEOUT = 5
+      REDIRECT_LIMIT = 10
+
+      attr_reader :link_responses
+
+      def initialize
+        @link_responses = {}
+      end
+
+      def check_links(links)
+        links.each do |link|
+          link_responses[link] = [fetch_status(link), Time.zone.now]
+        end
+        link_responses
+      end
+
+    private
+
+      def fetch_status(link)
+        begin
+          response = connection.get(URI.parse(link)) do |request|
+            request.options[:timeout] = TIMEOUT
+            request.options[:open_timeout] = TIMEOUT
+          end
+          response.status.to_s
+        rescue Faraday::ConnectionFailed
+          "Connection failed"
+        rescue Faraday::TimeoutError
+          "Timeout Error"
+        rescue Faraday::SSLError
+          "SSL Error"
+        rescue FaradayMiddleware::RedirectLimitReached
+          "Too many redirects"
+        rescue URI::InvalidURIError
+          "Invalid URI"
+        rescue => e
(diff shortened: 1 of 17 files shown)


lib/local-links-manager/check_links/link_checker.rb
```diff
@@ -0,0 +1,52 @@
...
+          "Too many redirects"
+        rescue URI::InvalidURIError
+          "Invalid URI"
+        rescue => e
+          e.class.to_s
+        end
+      end
+
+      def connection
+        @connection ||= Faraday.new(headers: { accept_encoding: 'none' }) do |faraday|
```
h-lame, line 45: Do we have some documentation about why we've set this header?  It seems unusual to have done so.

spec/spec_helper.rb
```diff
@@ -90,3 +90,5 @@
   Kernel.srand config.seed
 =end
 end
+
+require 'timecop'
```
h-lame: Probably better to either put this at the top of the file near the other require's, or put it in a file in spec/support.  It's usually a good idea to try and keep spec_helper.rb as untouched as possible so that when upgrading the diff between our version and the new one that the new version of rspec wants to use is minimal (e.g. only the changes that rspec wants to make).

lib/local-links-manager/check_links/link_checker.rb
davidbasalla, line 45: We ran into a URL which we could only get to work by setting this option. I'll try to find out what the problem was and give an update

davidbasalla, line 45: So when we make a GET request to http://councillors.dudley.gov.uk, (without the encoding option) we get the following error:

```
Zlib::DataError: unknown compression method
```

Googling reveals this (https://github.com/lostisland/faraday/issues/120) and suggests to set that option as a workaround. It sounds like it's caused by combination of a Ruby bug and what the server is sending back? I guess there's a question here of whether we want to support this kind of thing or whether we want to flag this as an error?

---

Update: I've added more details in the commit message about this

spec/spec_helper.rb
davidbasalla: Thanks, have moved it to a separate file

h-lame:
This is good stuff!

PR #29, opened 2016-07-05
Add cron job for link checker

brenetic:
- Added Redis for doing distributed lock so that the rake task only runs on one machine.
- Added Whenever gem to schedule cron job to run the link checker task daily.
- Added notify_passive_check script to notify Icinga of success/failure.
- Moved all other import tasks to whenever (see: alphagov/govuk-puppet#4591)

[Trello card](https://trello.com/c/KBT9RC9q/406-check-the-status-code-of-links)

[Related PR](https://github.com/alphagov/govuk-puppet/pull/4587)

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -20,4 +20,10 @@ module Services
       Redis.new(redis_config)
     end
   end
+
+  def self.icinga_check(service_desc, code, message)
+    unless Rails.env.development?
+      `/usr/local/bin/notify_passive_check "#{service_desc}" #{code} "#{message}"`
+    end
+  end
 end
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -1,3 +1,4 @@
+require 'redis'
 require 'gds_api/mapit'
 
 module Services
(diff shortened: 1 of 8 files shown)


lib/services.rb
```diff
@@ -1,10 +1,27 @@
...
+        port: ENV["REDIS_PORT"] || 6379,
+        namespace: "local-links-manager",
+      }
+
+      Redis.new(redis_config)
+    end
+  end
+
+  def self.icinga_check(service_desc, code, message)
+    `notify_passive_check "#{service_desc}" #{code} "#{message}"`
```
alexmuller: You might want to use the full path here, I'm not sure what the deploy user's `$PATH` looks like.

brenetic:
The notification to Icinga will currently run in dev whenever the rake task is run. Do we not want to do that?

brenetic:
Having tested running the rake tasks in development, we definitely want to not talk to icinga in dev as it breaks.  Anyone got any betters ideas than adding a `unless Rails.env.development?` guard clause to the `Service.icinga_check`?

sihugh:
I think the guard clause is fine, and this stuff is really useful.

PR #30, opened 2016-07-07
Change LinkChecker class to check individual links

klssmith:
- Previously we could not see any output because the script was waiting until the very end before outputting.
- Changed LinkChecker class to check links one at a time.
- This lets the output of checked links be monitored earlier.
- Changed HomepageStatusUpdater and LinkStatusUpdater class to update links one at a time.

### DIFF
--- a/lib/local-links-manager/check_links/homepage_status_updater.rb
+++ b/lib/local-links-manager/check_links/homepage_status_updater.rb
@@ -12,8 +12,10 @@ module LocalLinksManager
       end
 
       def update
-        links_responses = url_checker.check_links(links)
-        update_links(links_responses)
+        links.each do |link|
+          link_response = url_checker.check_link(link)
+          update_link(link, link_response)
+        end
       end
 
     private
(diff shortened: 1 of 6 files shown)


brenetic:
Considering their similarities is it worth having `homepage_status_updater` and `link_status_updater` as a more generic class or module?

h-lame:
This looks exactly what was needed, good stuff @klssmith 👍

FWIW I have a similar mild concern along the same lines as @brenetic; the two status updaters are v. similar and could stand to have some duplication removed (either via a base class and inheritance or extracting another object and using composition).  That said, as part of the discussion that made us do this PR we also discussed a further change to use a separate table for the urls.  If we decide that we should also make this change (not now, but later) then we'd only have one status updater and it's not really worth resolving the duplication between the two existing ones knowing that one will go away later and we'll essentially undo the work.

I'd be happy to merge as-is, but appreciate we might want a tiny discussion about the duplication first.

sihugh:
Extracting the urls feels OK to me, especially as homepage_urls seem to creep into the service_interaction links on a regular basis.  I'm going to merge this, and we can argue anew when everyone's back in if necessary :-)

PR #31, opened 2016-07-12
Alert when missing service or interaction

sihugh:
We import data from a number of sources:
- local authorities
- services 
- interactions
- the allowed mapping of services/interactions
- links

Links are dealt with slightly differently because we will stop importing them at some point and become the sole source of the truth ourselves.

The other data sources are beyond our control to varying extents.  Things that we rely on may well disappear:
- Local authority boundaries change, new ones are created and some disappear altogether
- It's plausible that new service or interaction types will be introduced, some will be retired and the mappings between the two could change.

Because we can't see the future, we can't readily predict what our intentions would be for any of these events.  It's likely that the work that we'd have to do would not be limited to this application. Instead of cleaning up automatically, we'll just ensure that we find out about these occurrences as soon as possible.

This PR introduces Icinga alerts for import discrepancies for Local Authorities, Services and Interactions.

[Trello card](https://trello.com/c/KuN7jtEr/394-imports-should-delete-things-sensibly-5)

### DIFF
--- a/lib/local-links-manager/import/import_comparer.rb
+++ b/lib/local-links-manager/import/import_comparer.rb
@@ -1,0 +1,55 @@
+class ImportComparer
+  def initialize(import_type)
+    @import_type = import_type
+    @records_in_source = Set.new
+    @missing = Set.new
+  end
+
+  def add_source_record(record_key)
+    @records_in_source.add(record_key)
+  end
+
+  def check_missing_records(saved_records)
+    saved_records.each do |record|
+      record_key = yield(record)
+      unless @records_in_source.include? record_key
+        @missing.add(record_key)
+      end
+    end
+
+    notify_record_status
+  end
+
+private
+
+  def notify_record_status
+    @service_desc = "Import #{@import_type.pluralize} into Local Links Manager"
+
+    if @missing.empty?
+      confirm_records_are_present
+    else
+      alert_missing_records
+    end
+  end
+
+  def confirm_records_are_present
+    Services.icinga_check(@service_desc, true, "Success")
+  end
+
+  def alert_missing_records
+    Services.icinga_check(@service_desc, false, error_message(@missing))
+  end
+
+  def error_message(missing)
+    suffix = "no longer in the import source."
+    if missing.count == 1
+      "1 #{@import_type} is #{suffix}"
+    else
+      "#{missing.count} #{@import_type.pluralize} are #{suffix}\n#{list_missing(missing)}\n"
+    end
+  end
+
+  def list_missing(missing)
+    missing.to_a.sort.join("\n")
+  end
+end
--- a/lib/local-links-manager/import/interactions_importer.rb
+++ b/lib/local-links-manager/import/interactions_importer.rb
@@ -13,12 +13,21 @@ module LocalLinksManager
         new.import_records
       end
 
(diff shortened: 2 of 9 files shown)


lib/local-links-manager/import/interactions_importer.rb
```diff
@@ -15,16 +15,15 @@ def self.import
 
       def initialize(csv_downloader = CsvDownloader.new(CSV_URL, header_conversions: FIELD_NAME_CONVERSIONS))
         @csv_downloader = csv_downloader
-        @lgil_codes_in_csv = Set.new
+        @comparer = ImportComparer.new("interactions")
```
davidbasalla: For the `CsvDownloader` we used dependency injection, could do the same with `ImportComparer`?

lib/local-links-manager/import/import_comparer.rb
```diff
@@ -0,0 +1,32 @@
...
+      record_key = yield(record)
+      unless @records_in_source.include? record_key
+        @missing.add(record_key)
+      end
+    end
+
+    alert_missing_records
+  end
+
+  def alert_missing_records
```
davidbasalla: I think `alert_missing_records` and `error_message` could be a private methods

lib/local-links-manager/import/import_comparer.rb
```diff
@@ -0,0 +1,32 @@
+class ImportComparer
```
davidbasalla, line 1: I like this new class 👍 We could probably add some unit tests for it? I guess it's already being indirectly tested in the Importer class tests, but would be nice to have explicit tests for its interface

lib/local-links-manager/import/interactions_importer.rb
sihugh: Good point, just about to push this up, thanks.

spec/lib/local-links-manager/import/services_importer_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'rails_helper'
 require 'local-links-manager/import/services_importer'
+require 'services'
```
davidbasalla: Is this `require` statement needed here? We don't seem to do it for the Interactions specs

sihugh: Seems legit 👍

lib/local-links-manager/import/import_comparer.rb
sihugh, line 1: Agreed.

spec/lib/local-links-manager/import/interactions_importer_spec.rb
```diff
@@ -51,5 +51,51 @@
         LocalLinksManager::Import::InteractionsImporter.new(csv_downloader).import_records
       end
     end
+
+    context 'when an interaction is no longer in the CSV' do
```
davidbasalla: Now that we have a separate spec for the `ImportComparer` class, I'm wondering if we can remove these tests here, since they test the same thing (expecting that the `Services` module receives the method `icinga_check`)? I'm not 100% sure though so happy to chat in person tomorrow.

If we do leave them, we should add some specs for the `LocalAuthority` importer too.

davidbasalla:
Awesome! Merging

PR #32, opened 2016-07-13
Add Icinga status for start of link checker

sihugh:
We want to be a bit more clear about which server is running the link checker task. We were a bit confused this morning because the "freshold" (thanks @surminus) had been exceeded and we weren't sure what was happening.

cc @brenetic @davidbasalla


brenetic:
👍

PR #34, opened 2016-07-14
Alert us of empty homepage URLs

sihugh:
We rely on homepage URLs being present as they are the ultimate fall back for local_transaction URLs if no more specific link exists.  It is important that we have something present. This PR also adds a schedule for running the homepage_url import as it is only currently run manually.  The link checker should run after the import to tell us if the URLs are incorrect.

I've configured a freshness threshold in this PR: https://github.com/alphagov/govuk-puppet/pull/4643

### DIFF
--- a/lib/tasks/import/local_authorities.rake
+++ b/lib/tasks/import/local_authorities.rake
@@ -1,3 +1,4 @@
+require 'local-links-manager/distributed_lock'
 require 'local-links-manager/import/local_authorities_importer'
 require 'local-links-manager/import/local_authorities_url_importer'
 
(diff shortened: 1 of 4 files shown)


spec/lib/local-links-manager/import/local_authorities_url_importer_spec.rb
```diff
@@ -50,5 +50,51 @@ def stub_slug_csv_request(data)
       la_allerdale = LocalAuthority.find_by(snac: "16UB")
       expect(la_allerdale.homepage_url).to eq("https://www.allerdale.gov.uk")
     end
+
+    context "when any homepage URLs are blank or empty after an import" do
+      it "should alert Icinga" do
+        FactoryGirl.create(:local_authority, name: 'adur', snac: "45UB", homepage_url: nil, gss: "E07000223")
+        FactoryGirl.create(:local_authority, name: 'allerdale', snac: "16UB", homepage_url: nil, gss: "E07000026")
+        FactoryGirl.create(:local_authority, name: 'hogsmeade', snac: "WZRD", homepage_url: nil, gss: "M491C0000")
```
klssmith: Now that we're using sequences in our LocalAuthority factory, would it be worth simplifying the test setup here by not specifying things that we don't have to any more (e.g. the gss code)?

sihugh: But I did a special GSS code for Hogsmeade! :-(

sihugh: I've removed those bits now, thanks @klssmith

brenetic:
Worth `Rails.logger`'ing?

lib/tasks/import/local_authorities.rake
```diff
@@ -17,7 +18,22 @@ namespace :import do
...
      imported by running `import_authorities`"
     task add_urls: :environment do
-      LocalLinksManager::Import::LocalAuthoritiesURLImporter.import_urls
+      service_desc = "Check for blank homepage urls in local-links-manager"
+      LocalLinksManager::DistributedLock.new('import-homepages').lock(
+        lock_obtained: ->() {
+          begin
+            LocalLinksManager::Import::LocalAuthoritiesURLImporter.import_urls
+            # Flags nagios that this servers instance succeeded to stop lingering failures
+            LocalLinksManager::Import::LocalAuthoritiesURLImporter.alert_empty_urls(service_desc)
```
brenetic, line 30: Do we want to alert Icinga of success after these?

sihugh:
I was hoping that last night's rake tasks would have helped us work out if it was useful or not, but :+1:

brenetic:
👍

PR #35, opened 2016-07-14
Add logging to link-checker rake task

sihugh:
We've experienced a few unexpected freshness threshold alerts in Icinga.  We want to increase our knowledge of exactly what is happening, so we're adding logs on start and finish of the rake task.


brenetic:
👍

PR #37, opened 2016-07-15
Disable services that are not in the CSV

sihugh:
In addition to enabling services that are in the CSV, we want to ensure that we disable the others in our database.  This allows us to remove unsupported services by using the CSV without having to write migrations to support it.

At some point in the near future we will stop having to import this CSV as local links manager will control which services we support, not a CSV in Publisher.

### DIFF
--- a/lib/local-links-manager/import/enabled_service_checker.rb
+++ b/lib/local-links-manager/import/enabled_service_checker.rb
@@ -14,9 +14,15 @@ module LocalLinksManager
       end
 
       def enable_services
-        @csv_downloader.each_row do |row|
-          set_enabled_service(row)
-        end
+        @supported_lgsl_codes = Set.new
+
+        @csv_downloader.each_row { |row| @supported_lgsl_codes.add(row["LGSL"]) }
+
+        Service.all.each { |service| set_enabled_state(service) }
+
+        check_for_missing_services
+
+        log_result
       rescue CsvDownloader::Error => e
         Rails.logger.error e.message
       rescue => e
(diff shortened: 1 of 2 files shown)


lib/local-links-manager/import/enabled_service_checker.rb
```diff
@@ -25,14 +33,41 @@ def enable_services
...
 
-      def set_enabled_service(row)
-        service = Service.find_by(lgsl_code: row["LGSL"])
-        if service.nil?
-          Rails.logger.warn("'#{row['LGSL']}' is not an imported Service")
-        else
-          service.enabled = true
-          Rails.logger.info("'#{row['LGSL']}' enabled")
-          service.save!
+      def supported_lgsl_codes
```
davidbasalla: Could this be an instance attribute instead? Or does it need to be memoised?

lib/local-links-manager/import/enabled_service_checker.rb
```diff
@@ -15,8 +15,16 @@ def initialize(csv_downloader = CsvDownloader.new(CSV_URL))
 
       def enable_services
         @csv_downloader.each_row do |row|
```
davidbasalla: Personal preference so feel free to ignore: Could we use the one-liner versions (ie. `@csv_downloader.each_row { |row| supported_lgsl_codes.add(row["LGSL"]) }`) instead? Same for Line 21-23

davidbasalla:
Thanks @sihugh, this looks good, only a couple of minor comments. Thought we would have to do this as part of https://trello.com/c/G8GfEDcH/424-disable-services-in-local-links-manager-that-are-actually-licences

lib/local-links-manager/import/enabled_service_checker.rb
sihugh: Sure, why not! :-)

lib/local-links-manager/import/enabled_service_checker.rb
sihugh: I was trying to tidy up `enable_services` a bit, that's all.  No real preference either way.

sihugh:
Thanks @davidbasalla, I've made the suggested changes.

PR #38, opened 2016-07-18
Display link quality

brenetic:
[Trello card](https://trello.com/c/iiE3yD99/405-display-basic-link-quality-data-in-local-links-manager-pair-with-mia)

Display link status and last checked time
- We now show the status of all the interaction links and the time they
  were last checked on the page.
- Set the link status and time last checked back to nil if it is edited.
- If there is no link we just display the text "No link". Similarly if
  the link has not been checked "Link not checked" is displayed.

Display status and time for homepage url
- We now show the status of the link and the time it was last checked
  below the homepage url.
- We reset the status and time last checked to nil when a homepage url is
  edited.

Initially we added a number of methods to the view helpers, but eventually decided to go with the Presenter pattern (https://github.com/alphagov/local-links-manager/pull/38/commits/ec8b426b569b574eb791d64e7bed1d6d903211d5) as it allows for more object-oriented design and makes handling interactions and their links more convenient. Because of this change, the files change quite a bit during that commit (something to keep in mind when reviewing the first 2 commits). We add feature tests to test the new features, and unit tests to test the Presenters thoroughly.

Finally we change the Timecop behaviour to return after each test by default in the initializer rather than having to explicitly return after the tests that require it. We also change the test order in Rspec to random so that we reveal any order dependencies.

---

NOTE: Several times during the story we came across problems regarding how LocalAuthorities store their links. The `LocalAuthority` model has a `homepage_url`, `status` and `link_last_checked` fields. For Interactions we have the separate `Link` model which encapsulates those same attributes. So this would suggest that `LocalAuthority` could benefit from a similar model for its homepage. This will probably be addressed in https://trello.com/c/TiVpH2I4/331-consolidate-localauthority-homepage-url-and-link-url-behaviour

---
# Before

image
# After

image

image


sihugh:
I noticed that when a `homepage_url` had a `Timeout error` status then it looked a little weird.  The class was missing on the span to do the lozenge which made it (the lozenge and the text) invisible.  The "Checked about an hour ago" text was still visible, but was indented by the width of the lozenge span.  I'm not sure what the desired behaviour was here - not show anything?

screen shot 2016-07-18 at 17 21 20

I was going to try some of the other statuses, and whether the same sort of thing happens for the interaction urls.

sihugh:
I deployed the branch to integration to check - the interaction url status displays fine with a `Timeout Error`, but the homepage url status has the same colour as the background.  The `label` class sets the text color to `fff`, so we might need a bit of work there.

screen shot 2016-07-19 at 10 11 53

app/models/link.rb
```diff
@@ -1,4 +1,6 @@
 class Link < ActiveRecord::Base
+  after_update :reset_time_and_status, if: :url_changed?
```
sihugh, line 2: We're adding more duplication with the LocalAuthority model.  I think this is OK, but I've added a Trello card to refactor this behaviour out -  https://trello.com/c/TiVpH2I4/331-consolidate-localauthority-homepage-url-and-link-url-behaviour

We could do this now?  Possibly need to talk it out in stand up?

brenetic, line 2: Looks like @klssmith @davidbasalla and @emmabeynon are going to implement a presenter to take care of this. They may well refactor the above at the same time?

brenetic:
`url_status_presenter`? instead of `url_status_presentation`?

brenetic:
👍

app/models/link.rb
davidbasalla, line 2: Thanks for writing up that story @sihugh, we've come up against this problem a few times during this story. We didn't attempt to fix it in this story, as there's a few considerations to uh consider about how to best model the URL related fields for a LocalAuthority (see comment above in description)

davidbasalla:
@brenetic About your comment regarding "`url_status_presenter?` instead of `url_status_presentation?`", we went with the latter because it's a module and not an actual Presenter so were hoping to avoid confusion. We were wondering whether it's more standard to suffix a module with 'able' but didn't find a conclusive answer online.

sihugh:
💊  👍

PR #39, opened 2016-07-19
Correct enabled service checker

sihugh:
We were incorrectly expecting the enabled service CSV file to provide us with
numeric values (when they are strings).  This threw out the comparator which
enables/disables the services we're interested in resulting in us disabling all
of the services.

We've fixed the tests to be more reflective of reality and now compare strings
instead.


brenetic:
⚡

PR #40, opened 2016-07-21
Update lozenge styling and fix Link Status display

davidbasalla:
This commit adds extra space to the right of the lozenges and displays
the Link Status as grey when there is no link.

We had to move both opening and closing <span> elements onto one
line to get the correct spacing. The following link displays the effect in
detail: http://stackoverflow.com/a/25879479
# Before

screen shot 2016-07-21 at 10 56 10
# After

screen shot 2016-07-21 at 10 56 17


brenetic:
💭

PR #41, opened 2016-07-21
Generate CSVs for homepage link status and interaction link status

issyl0:
- This generates CSVs of link check statistics when people access `/check_homepage_links_status.csv` and `/check_links_status.csv`, so we can do reporting into the numbers of broken links etc.
- Trello: https://trello.com/c/BCMg9RTo/426-get-the-data-for-broken-links-2

(Paired with @klssmith.)

### DIFF
--- a/app/presenters/link_check_csv_presenter.rb
+++ b/app/presenters/link_check_csv_presenter.rb
@@ -1,0 +1,21 @@
+require 'csv'
+
+class LinkCheckCSVPresenter
+  def self.homepage_links_status_csv
+    CSV.generate do |csv|
+      csv << %w(status count)
+      LocalAuthority.group(:status).count.each do |key, value|
+        csv << [key || "nil", value]
+      end
+    end
+  end
+
+  def self.links_status_csv
+    CSV.generate do |csv|
+      csv << %w(status count)
+      Link.enabled_links.group(:status).count.each do |key, value|
+        csv << [key || "nil", value]
+      end
+    end
+  end
+end
--- a/spec/controllers/links_controller_spec.rb
+++ b/spec/controllers/links_controller_spec.rb
@@ -2,6 +2,7 @@ require 'rails_helper'
 
 RSpec.describe LinksController, type: :controller do
   before do
+    login_as_stub_user
     @local_authority = FactoryGirl.create(:local_authority, name: 'Angus')
     @service = FactoryGirl.create(:service, label: 'Service 1', lgsl_code: 1)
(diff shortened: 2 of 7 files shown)


app/presenters/link_check_csv_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+      LocalAuthority.group(:status).count.each do |key, value|
+        csv << [key || "nil", value]
+      end
+    end
+  end
+
+  def self.links_status_csv
+    CSV.generate do |csv|
+      csv << %w(status count)
+      Link.group(:status).count.each do |key, value|
```
jennyd: Could we scope this to only include links for services which are enabled? All other links aren't checked at all, so it's harder to work out how many links should be checked but haven't been yet for some reason (and so still have `nil` as their status) if we include all the links we'll never check.

sihugh: I agree although it does slightly beg the question why we hold on to old links for services that we don't care about and don't check.  Should we either check them or get rid of them?  (That's not really to do with this PR though, sorry)

jennyd: We may enable other services in future, and having the old links in the database already would give us a starting point for those. The checking takes a while, though, and would take much longer if we included all the disabled service links - we don't need quality data until we want to start using them.

spec/controllers/links_controller_spec.rb
```diff
@@ -2,26 +2,38 @@
...
       expect(response).to have_http_status(302)
       expect(flash[:danger]).not_to be_present
     end
   end
+
+  describe 'GET homepage_links_status_csv' do
+    it "retrieves HTTP success" do
+      get :homepage_links_status_csv
+      expect(response).to have_http_status(:success)
+    end
```
sihugh, line 31: We have tests for generating the CSV and ensuring that we can get a successful response from the route, but nothing that knits the two together.  Maybe we could have a feature spec that ensures that we serve a CSV when you hit this endpoint and it has the expected columns?

app/presenters/link_check_csv_presenter.rb
sihugh: I understand, it just makes me feel slightly icky having it hanging around uncared for.  I'll worry more about it in the Autumn. :-)

issyl0: Updated.

sihugh:
🎫

PR #42, opened 2016-07-22
[DISCUSS] Enable show route for a JSON request for a specific Link

davidbasalla:
## For discussion about LocalLinksManager API, do not merge

We enable the show route for nested Interactions, so the link for the specific ServiceInteraction can be requested.
The request would look like:

`GET /local_authorities/bolton/services/412/interactions/8.json`

The most contentious part of this implementation is probably that we have to convert from LGSL and LGIL codes to slugs in the controller. Frontend will make the request and only knows about LGSL and LGIL codes. However the routes we have set up in LocalLinksManager currently all work with slugs. That's why we need to convert the params hash in the controller to convert the numbers into relevant slugs.

### DIFF
--- a/app/controllers/interactions_controller.rb
+++ b/app/controllers/interactions_controller.rb
@@ -1,4 +1,27 @@
 class InteractionsController < ApplicationController
+  def show
+    sanitize_service_slug_param
+    sanitize_interaction_slug_param
+
+    @authority = LocalAuthorityPresenter.new(LocalAuthority.find_by_slug!(params[:local_authority_slug]))
+    @service = Service.find_by_slug!(params[:service_slug])
+    @interactions = @service.interactions
+    @link = presented_links.detect { |l| @interactions.first.service_interactions.first.id == l.service_interaction.id }
+
+    # Should have a method like?
+    # links = @authority.links.for_service_interaction(@service, @interaction)
+
+    response = {
+      "local_authority"=>{
+        "name"=>@authority.name,
+      },
+      "local_interaction"=>{
+        "url"=>@link.url
+      }
+    }
+    render :json => response
+  end
+
   def index
     @authority = LocalAuthorityPresenter.new(LocalAuthority.find_by_slug!(params[:local_authority_slug]))
     @service = Service.find_by_slug!(params[:service_slug])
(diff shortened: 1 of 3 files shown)


jennyd:
I agree that that aspect is contentious :fire: :fire_engine: :ocean: I think it's generally better not to modify `params` if we can avoid it [citation needed - @h-lame?]. Using both slugs and numbers on the same route feels like something that will cause us problems at some point.

Using a RESTful-looking URL suggests that we're going to return exactly the requested interaction resource, whereas actually we'll return whatever link the fallback logic determines (often for a different interaction), plus the details about the LA that Frontend needs to render the page. Using query params would imply that less strongly to me, eg something like `/local-link?la_slug=bolton&lgsl=412&lgil=8`.

jennyd:
We should consider whether we want the API route to be behind authentication or not. If it is, then every request to a local transaction results page that isn't served from a cache will trigger a request to Signon, and if Signon is down then it will error. Imminence also has both an admin interface and an API used by Frontend, and it [doesn't require authentication for the API](https://github.com/alphagov/imminence/blob/2c859bd064092fc9d2577f650ede693142f14c9a/app/controllers/places_controller.rb).

davidbasalla:
@jennyd We just created https://github.com/alphagov/local-links-manager/pull/43 as an alternative, which does what you suggested, and we agree that it probably feels more appropriate to go with that solution (so not the one presented here). Changing the params from numeric to slugs, and using the RESTful approach, when the resources we return might actually be different doesn't feel great.

davidbasalla:
@jennyd good point about the authentication, we were also talking about that. Sounds like the way Imminence does it is the way to go?

h-lame:
FWIW: as an approach #43 seems like the correct route, so another :-1: on this from me.  

Responding to the [citation needed] from @jennyd, it's correct that we should avoid mutating values in `params` if we can get away with it.  That said it's not totally forbidden, sometimes it makes sense to do some type-coercion, or to add some default values, or even to break up an existing value into a bunch of other values.  The needs of your HTML form interface and the needs of your domain modelling or business logic aren't always in sync and I've always seen it as the controller's job to translate between the web request and the business logic.  Rails somewhat muddies this interpretation because the framework already turns the raw GET query string or POST body into a hash before you get to your action, and then lets ActiveRecord do some type-coercion of those params when writing to the db, but I think the principle still stands.  I'd always prefer to add new values to `params` over mutating existing ones if I could.

davidbasalla:
Closing as this PR was for discussion only, and https://github.com/alphagov/local-links-manager/pull/44 represents the actual implementation of the API endpoint.

PR #43, opened 2016-07-22
[DISCUSS] Add new api route and controller

davidbasalla:
## For discussion about LocalLinksManager API, do not merge

This is an alternative approach to https://github.com/alphagov/local-links-manager/pull/42

We add a separate controller to handle JSON requests for the API. We add a method to return local interaction details and find the related data through query params that are passed along with the request. 

A lot more logic would have to be added to handle the fallback logic for local interactions, but again just a broad outline.

A future 'local_authority' method could be added here for handling requests for 'Find my local council'.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -1,0 +1,28 @@
+class ApiController < ApplicationController
+  def service_interaction_for_local_authority
+    @authority = LocalAuthority.find_by_slug!(params[:authority_slug])
+    @service = Service.find_by(lgsl_code: params[:lgsl])
+    interactions = @service.interactions
+    @link = links.detect { |l| interactions.first.service_interactions.first.id == l.service_interaction.id }
+
+    render json: local_interaction_response
+  end
+
+private
+
+  def local_interaction_response
+    {
+      "local_authority"=>{
+        "name"=>@authority.name,
+      },
+      "local_interaction"=>{
+        "url"=>@link.url
+      }
+    }
+  end
+
+  def links
+    @_links ||= @authority.links.for_service(@service)
+  end
+end
+
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,6 +4,10 @@ Rails.application.routes.draw do
 
   get '/healthcheck', to: proc { [200, {}, ['OK']] }
 
+  get '/api/local_interaction_details', to: 'api#service_interaction_for_local_authority'
+  get '/api/local_authorities', to: 'api#local_authority'
+
+
   resources "local_authorities", only: [:index, :edit, :update], param: :slug do
     resources "services", only: [:index], param: :slug do
       resources "interactions", only: [:index], param: :slug do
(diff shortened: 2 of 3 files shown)


jennyd:
I think this approach is clearer than #42, both in terms of REST and also by separating out the temporary API functionality from the permanent backend interface.

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    render json: local_interaction_response
+  end
+
+private
+
+  def local_interaction_response
+    {
+      "local_authority"=>{
+        "name"=>@authority.name,
```
jennyd, line 16: We'd also need the homepage URL here.

app/controllers/api_controller.rb
```diff
@@ -0,0 +1,28 @@
+class ApiController < ApplicationController
+  def service_interaction_for_local_authority
+    @authority = LocalAuthority.find_by_slug!(params[:authority_slug])
+    @service = Service.find_by(lgsl_code: params[:lgsl])
+    interactions = @service.interactions
+    @link = links.detect { |l| interactions.first.service_interactions.first.id == l.service_interaction.id }
```
jennyd, line 6: ...plus the full fallback logic, of course :) :fallen_leaf:

app/controllers/api_controller.rb
davidbasalla, line 16: What would we use the homepage URL for? We were thinking that we only need the homepage in case no valid LGIL could be found, in which case we would return the homepage url as the local_interaction url.

jennyd, line 16: Frontend needs to know whether the URL is for an interaction or the homepage, because we use [different text](https://github.com/alphagov/frontend/blob/3b12a066f0787f9a926912becf04a8b440792f12/config/locales/en.yml#L38) on the page, eg https://www.gov.uk/pay-council-tax/dundee

app/controllers/api_controller.rb
davidbasalla, line 6: Actually, I think this code wouldn't work properly at the moment as we're not targeting the right Interaction (defined by `param[:lgil]`). General concept still stands though :)

issyl0:
I also prefer this to #42 - it's more obvious what it does rather than being mixed up in the standard InteractionsController that serves the main app.

app/controllers/api_controller.rb
davidbasalla, line 16: Just mentioned [this](https://github.com/alphagov/frontend/pull/977#discussion_r72037802) in the Frontend PR. Maybe we shouldn't be combining the two types of data (local authority data/local interaction data) in a single route/request. Instead could have 2 separate routes, one for local authority data and one for local interaction data. Frontend would have to make 2 requests to get the appropriate data.

h-lame:
I echo existing comments about the approach in general and specifics. :+1:

davidbasalla:
Closing as this PR was for discussion only, and https://github.com/alphagov/local-links-manager/pull/44 represents the actual implementation of the API endpoint.

PR #44, opened 2016-07-27
Add link api

davidbasalla:
For https://trello.com/c/U31MOcCC/396-2-of-4-add-the-api-endpoint-to-local-links-manager-3

We add an api endpoint for link details, expecting a local authority slug, lgsl and optionally lgil as query parameters to the route `/api/link`. We extract the logic for determining the correct link (including fallback logic) to a separate class which is easier to unit test and could be reused in future.

We also switch back to using `pry-rails` for debugging as we were getting Segmentation faults when using `binding.pry` in our rspec tests.

Mobbed with @emmabeynon, @brenetic and @h-lame

### DIFF
--- a/lib/local-links-manager/link_resolver.rb
+++ b/lib/local-links-manager/link_resolver.rb
@@ -1,0 +1,41 @@
+module LocalLinksManager
+  class LinkResolver
+    def initialize(authority, service, interaction = nil)
+      @authority = authority
+      @service = service
+      @interaction = interaction
+    end
+
+    def resolve
+      if @interaction
+        link_for_interaction
+      else
+        fallback_link
+      end
+    end
+
+  private
+
+    def link_for_interaction
+      @authority.links.find_by_service_and_interaction(@service, @interaction)
+    end
+
+    def fallback_link
+      if service_links_ordered_by_lgil.count == 1
+        service_links_ordered_by_lgil.first
+      else
+        link_with_lowest_lgil_but_not_providing_information_lgil
+      end
+    end
+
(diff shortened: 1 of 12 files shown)


lib/local-links-manager/link_resolver.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+    end
+
+  private
+
+    def link_for_interaction
+      @authority.links.find_by_service_and_interaction(@service, @interaction)
+    end
+
+    def fallback_link
```
sihugh, line 23: This is much clearer than before :+1:

spec/presenters/link_api_response_presenter_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+describe LinkApiResponsePresenter do
+  describe '#present' do
+    let(:authority) { FactoryGirl.build(:local_authority) }
+    let(:presenter) { described_class.new(authority, link) }
```
sihugh, line 6: It feels a little surprising to specify a thing without all its dependencies (`link`) being defined.  Is this a pattern I should be using?

sihugh:
Looks good :+1:

spec/presenters/link_api_response_presenter_spec.rb
brenetic, line 6: If I remember @h-lame correctly the `presenter` is called on line 27 and at that point `link` is available from line 9. Using outside this context it would indeed break.

PR #45, opened 2016-07-27
Fix homepage link status when blank

emmabeynon:
[Trello Card](https://trello.com/c/tibzgu4G/437-fix-homepage-link-status-when-we-don-t-have-a-homepage)

This allows the correct information to be displayed when a Local Authority homepage url is set to nil or an empty string.  Previously, the page displayed 'Link not checked' when a URL was not present.  We now display 'No link' and no last checked time.  Additionally, we found a bug where the link checker was checking blank homepage URLs and returning an 'Invalid URI' error.  This has now been fixed, and blank homepage URLs are no longer checked.

Paired with @klssmith

### DIFF
--- a/lib/local-links-manager/check_links/homepage_status_updater.rb
+++ b/lib/local-links-manager/check_links/homepage_status_updater.rb
@@ -21,7 +21,7 @@ module LocalLinksManager
     private
 
       def links
-        table.distinct.pluck(column)
+        table.distinct.where.not(column => '').pluck(column)
       end
 
       def update_link(link, link_response)
--- a/app/presenters/local_authority_presenter.rb
+++ b/app/presenters/local_authority_presenter.rb
@@ -8,4 +8,8 @@ class LocalAuthorityPresenter < SimpleDelegator
   def homepage_status
     homepage_url.blank? ? 'No link' : status_description
   end
+
+  def homepage_link_last_checked
+    homepage_url.blank? ? '' : last_checked
+  end
 end
--- a/app/presenters/local_authority_presenter.rb
+++ b/app/presenters/local_authority_presenter.rb
@@ -4,4 +4,8 @@ class LocalAuthorityPresenter < SimpleDelegator
   def homepage_button_text
     homepage_url.blank? ? 'Add link' : 'Edit link'
   end
+
+  def homepage_status
+    homepage_url.blank? ? 'No link' : status_description
+  end
 end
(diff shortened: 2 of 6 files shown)


lib/local-links-manager/check_links/homepage_status_updater.rb
```diff
@@ -21,7 +21,7 @@ def update
     private
 
       def links
-        table.distinct.pluck(column)
+        table.distinct.where.not(column => '').pluck(column)
```
davidbasalla, line 24: Only a thought, the commit message mentions that this method will only retrieve links that are not set to nil and that are not set to an empty string. The added code sort of looks like it only checks for empty strings (rather than something like `.where.not(column => [nil, ''])`. However the added test seems to demonstrate that it also works for homepages that are set to nil, so it's probably fine? I'm assuming this is due to the fact that `nil` in Postgres is stored as an empty string... I might have to read up on this :)

# 

Update:

Ran these queries in the console to see what the SQL would look like:

`.where.not(column => nil)` creates the following SQL:
`SELECT DISTINCT "local_authorities"."homepage_url" FROM "local_authorities" WHERE ("local_authorities"."homepage_url" IS NOT NULL)`

`.where.not(column => ‘’)` creates the following SQL:
`SELECT DISTINCT "local_authorities"."homepage_url" FROM "local_authorities" WHERE ("local_authorities"."homepage_url" != $1)  [["homepage_url", "”]]`

`.where.not(column => [‘’, nil])` creates the following SQL:
`SELECT DISTINCT "local_authorities"."homepage_url" FROM "local_authorities" WHERE (NOT (("local_authorities"."homepage_url" = '' OR "local_authorities"."homepage_url" IS NULL)))`

However performing `.where.not(column => ‘’)` does indeed exclude both `nil` and empty string values, so your solution is good!

davidbasalla:
Good stuff! 👍

lib/local-links-manager/check_links/homepage_status_updater.rb
emmabeynon, line 24: We thought there might be a nice Active Record method that would handle both nil and empty strings, but this was the neatest solution we found :)

h-lame, line 24: This _might_ be database dependent as I think mysql might treat `''` and `null` differently.  I'd need to check it against a mysql db - but happily we're not using mysql so no need to worry.  Just pointing it out for future reference.  The good thing here is that we have the tests that explain what we actually want so if we did switch dbs we'd spot the failure and change the implementation.  Good stuff!

PR #47, opened 2016-08-02
Add a favicon in the same style as other apps

issyl0:
- In Integration with this branch deployed, the favicon looks like:

screenshot 2016-08-02 12 14 26


h-lame:
I thought we'd need loads of different sizes, but transition doesn't so this looks good. :+1:

sihugh:
I wish we'd had `LOL` :-(

PR #49, opened 2016-08-03
Return LocalAuthority details for missing ServiceInteraction

davidbasalla:
When we request a LGSL and LGIL combination from the local links API which does not exist, we should still return the LocalAuthority details. Previously only a 404 and and an empty hash was returned, which meant Frontend was serving its "Sorry, we are experiencing technical difficulties" page. Instead Frontend should show the homepage for the given Service and LocalAuthority.

This fix was prompted by visiting '/report-stray-dog' which is LGSL 432 and LGIL 0. This is a ServiceInteraction that we do not support in LocalLinksManager. We have a story in the Custom Team backlog about coming up with a fix for this missing ServiceInteraction -> https://trello.com/c/y79vHOmC/330-investigate-local-transactions-for-lgsl-432-dog-wardens

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -19,7 +19,7 @@ private
   end
 
   def missing_objects?
-    authority.nil? || service.nil? || service_interaction_required_but_not_found?
+    authority.nil? || service.nil?
   end
 
   def authority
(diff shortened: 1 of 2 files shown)


h-lame:
Edgecases are the worst. :+1:

PR #51, opened 2016-08-03
Update specs to use mapit gds api adapter test helpers

davidbasalla:
Since we're using the gds api adapter for MapIt, we should use the test helpers that are provided for MapIt as well. It will keep us in sync with any future work related to MapIt in the gds-api-adapters gem.

### DIFF
--- a/spec/lib/local-links-manager/import/local_authorities_importer_spec.rb
+++ b/spec/lib/local-links-manager/import/local_authorities_importer_spec.rb
@@ -9,24 +9,14 @@ describe LocalLinksManager::Import::LocalAuthoritiesImporter do
     File.expand_path("fixtures/" + file, File.dirname(__FILE__))
   end
 
-  def stub_mapit_request(data)
-    stub_request(:any, mapit_url).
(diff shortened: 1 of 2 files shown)


spec/lib/local-links-manager/import/local_authorities_importer_spec.rb
```diff
@@ -202,24 +191,24 @@ def stub_mapit_request(data)
...
-                "type_name": "County council",
-                "generation_low": 1,
-                "country_name": "England",
-                "type": "CTY"
+              "parent_area": nil,
+              "generation_high": 1,
+              "all_names": {},
+              "id": 1724,
+              "codes": {
+                  "ons": "11",
```
h-lame: While you're de-indenting, you could de-indent these by two spaces too.

h-lame:
I'm into it :+1:

davidbasalla:
thanks @h-lame, fixed the indentation!

brenetic:
LGTM FFTM

PR #54, opened 2016-08-11
Remove links and authority url imports

davidbasalla:
We want to stop importing data from LocalDirectGov's site as we're transitioning away from that site. This PR stops the scheduled imports of local links and local authority homepage URLs. All related code is removed as well.

For https://trello.com/c/t820ZpYU/452-delete-ldg-import-to-local-links-manager-and-related-code-2

NOTE: Not to be merged/deployed before we officially transition from LocalDirectGov. Should be merged along with https://github.com/alphagov/govuk-puppet/pull/4761


brenetic:
Reminder to remove rake task instructions from README then 👍 MWA

brenetic:
README as been updated. This is read to be merged when we require it.

PR #55, opened 2016-08-11
Export local authority links and associated data to a CSV

sihugh:
- Add class, tests and route to export all links to csv.
- Add rake task.
- The plan is to put these on data.gov.uk and for them to be as
  close to the old LDG CSVs as possible so people can still use
  them.
- Add a scheduled job to run the export rake task daily at 3am
- There is a symlink from `shared/data` to `public/data` inside the current release on each backend box. This done in `govuk-app-deployment` to persist the data between releases.

We don't want to use a distributed lock on this task as we need the file to
exist on each host.

[Trello card](https://trello.com/c/dFFQRlsp/448-generate-local-links-csv-3)

Mobbed with @brenetic @issyl0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,8 @@
 /tmp
 /coverage
 
-.DS_Store
+# Ignore /data, 'cause that's where we're storing the data exports.
+public/data/*
+!public/data/.keep
 
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ Import all the links for each local authority:
 
 `bundle exec rake import:links:import_all`
 
+### Exporting Local Authority links to services
+
+Exports are run daily and produce a `links.csv` file in the `public/data` directory. This is publicly available data found at `https://local-links-manager.publishing.service.gov.uk/data/links.csv`
+
+`bundle exec rake export:links:all`
+
 ### Example API output
 
 **Endoint for local authorities**
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -13,6 +13,10 @@ class LinksController < ApplicationController
     send_data data, filename: filename
   end
 
+  def exported_links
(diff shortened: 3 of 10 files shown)


davidbasalla:
I'm wondering if this would be more future proof if we went for using ActiveRecord models when writing out the data to CSV rather than a big SQL query. If in future our database table structure changes, this SQL statement may have to be updated as well, which is arguably harder than dealing with (primarily the `Link`) Active Record objects. Just a thought though, curious what other people think.

sihugh:
We started by using ActiveRecord but found that it didn't work quite how we thought and ended up making a load of individual queries (which took significantly longer - ~9 minutes as opposed to ~ 10 secs for this query).

We took inspiration from [Transition](https://github.com/alphagov/transition/blob/master/lib/transition/export/foi_response.rb) so assumed this was not inappropriate.

issyl0:
@davidbasalla We tried the ActiveRecord equivalent of this query, but it was very slow (~10 minutes) compared to the ~10 seconds the raw SQL takes (both running on integration). This raw SQL query [follows Transition's lead for exporting things](https://github.com/alphagov/transition/blob/master/lib/transition/export/foi_response.rb).

h-lame:
We could do a combo of both and use the ActiveRelation API to build the SQL query but still use the COPY / CSV features of Postgres to run the query and avoid instantiating all the objects that would slow it down when using AR.

davidbasalla:
Yeah, as an academic exercise it would be interesting to see if the slow performance was caused by loads of SQL queries (something which should be solvable with using `.includes`) or the object instantiation. As personal preference, I'd be happier with less raw SQL even with an 'acceptable' hit to performance. Let me know if you want to try what @h-lame suggested above, would be interesting to see how that works...

brenetic:
@davidbasalla @h-lame  Hybrid version in effect. 40% slower than raw SQL dump, but still a lot faster than the first approach.

h-lame:
Good stuff.  Merging.

PR #58, opened 2016-08-16
Remove authentication for csv export

brenetic:
We need the `/links-export` endpoint to open to the public therefore we
are skipping signon authentication for it.

[Trello card](https://trello.com/c/dFFQRlsp/448-generate-local-links-csv-3)


davidbasalla:
👍 💥

PR #59, opened 2016-08-22
Manually unlock locked rake tasks

klssmith:
There have been instances where rake tasks have not completed for
for some reason leaving the Distributed Lock for that task in
place. It is then necessary to wait for the lifetime of the lock
to end before being able to run task again.

This rake task adds the ability for the lock to be removed for the
given task passed in.

`Rails.env` was removed from the key as it was felt to be
superfluous since all rake tasks run in the same environment.

[Trello card](https://trello.com/c/OtTKMe3H/465-enable-unlocking-of-scheduled-tasks-2-pair-with-katie-pls)

### DIFF
--- a/lib/local-links-manager/distributed_lock.rb
+++ b/lib/local-links-manager/distributed_lock.rb
@@ -2,20 +2,34 @@ require 'redis-lock'
 
 module LocalLinksManager
   class DistributedLock
+    attr_accessor :lock_name, :redis_lock
+
+    APP = "local-links-manager"
     LIFETIME = (60 * 60) # seconds
 
     def initialize(lock_name)
       @lock_name = lock_name
+      @redis_lock = Redis::Lock.new(Services.redis, "#{APP}:#{lock_name}", owner: APP, life: LIFETIME)
     end
 
     def lock(lock_obtained:, lock_not_obtained:)
-      Services.redis.lock("local-links-manager:#{Rails.env}:#{@lock_name}", life: LIFETIME) do
-        Rails.logger.debug('Successfully got a lock. Running...')
-        lock_obtained.call
-      end
+      redis_lock.lock
+      Rails.logger.debug('Successfully got a lock. Running...')
+      lock_obtained.call
     rescue Redis::Lock::LockNotAcquired => e
-      Rails.logger.debug("Failed to get lock for #{@lock_name} (#{e.message}). Another process probably got there first.")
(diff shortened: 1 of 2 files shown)


lib/local-links-manager/distributed_lock.rb
```diff
@@ -2,20 +2,34 @@
...
         lock_obtained.call
       end
     rescue Redis::Lock::LockNotAcquired => e
-      Rails.logger.debug("Failed to get lock for #{@lock_name} (#{e.message}). Another process probably got there first.")
+      Rails.logger.debug("Failed to get lock for #{lock_name} (#{e.message}). Another process probably got there first.")
       lock_not_obtained.call
     end
+
+    def unlock
+      Redis::Lock.new(Services.redis, "#{APP}:#{lock_name}", owner: "#{APP}").unlock
```
h-lame: I guess we're not using `Services.redis.unlock` here because that doesn't let us specify the `owner` param (which is why it wasn't unlocking across processes?).  It'd be useful to mention this in the commit message, or a comment as it's important detail we might otherwise forget and break.

I might even consider changing how we obtain the lock to avoid the `Services.redis.lock` method and go long-hand with `Redis::Lock.new(Services.redis, "#{APP}:#{lock_name}", owner: APP).lock` so that our `lock` and `unlock` methods are more similar.   Perhaps with a comment like: avoid convenience methods on Services.redis to give us more options.  Not hugely convinced by this though, feel free to disregard this idea.

h-lame: ps.  Where you have `"#{APP}"` you can just use `APP` as it's already a string, no need to interpolate to get a new one.

lib/local-links-manager/distributed_lock.rb
```diff
@@ -2,20 +2,34 @@
...
+      Rails.logger.debug("Successfully unlocked #{lock_name}")
+    rescue StandardError => e
+      Rails.logger.error("Failed to unlock #{lock_name}\n#{e.message}")
+    end
+
+    def locked?
+      Redis::Lock.new(Services.redis, "#{APP}:#{lock_name}").locked?
```
h-lame: Not sure how `Redis::Lock` works, but will this need the `owner: APP` argument added?  If so, maybe we can move that into a helper method so that all the places that need to build a `Redis::Lock` do the same thing.

brenetic:
@h-lame  Thanks for the comments. Have updated.

h-lame:
Good stuff!

PR #60, opened 2016-08-31
Serve links CSV file from NGINX

sihugh:
We wanted to ensure that the links CSV file was served with the correct content type (`text/csv`).  The govuk-puppet NGINX configuration did not have this mime type defined and we were discouraged from modifying this at the time.  This led us down the path of defining the header in a Rails controller which works fine.

We've changed our approach back to the original idea due to the links CSV file being quite large.  It would be easy for us to tie up all our worker processes serving this file.

We've already [amended the NGINX config](https://github.com/alphagov/govuk-puppet/pull/4766) and deployed the change (it's a common file type after all and possibly should have been there in the first place).

We've expanded on the README section for the links CSV.

There are a couple of other minor naming changes in specs to hopefully increase clarity.

[Trello card](https://trello.com/c/pDCppZGk/479-make-nginx-handle-the-exported-links-csv-for-local-links-manager)


brenetic:
💭

PR #61, opened 2016-09-06
Check URL status and time for updated and new Links

emmabeynon:
- [Trello card](https://trello.com/c/Bh7piC4k/480-fix-the-status-of-links-that-have-recently-been-edited-and-the-url-was-checked-in-the-previous-link-check-run)
- When a link's URL is updated, or a new link is created, we now check if its URL has been recently checked.  If it has, then the link's status and link last checked time will be updated with the most recent data.
- We decided not to implement this functionality for Local Authority homepage URLs, as they are unique to each Local Authority and we wouldn't expect there to be duplicates in the Local Authority database.

Mobbed with @brenetic & @klssmith

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,5 +1,6 @@
 class Link < ActiveRecord::Base
-  after_update :reset_time_and_status, if: :url_changed?
+  before_update :set_time_and_status_on_updated_link, if: :url_changed?
+  before_create :set_time_and_status_on_new_link
 
   belongs_to :local_authority
   belongs_to :service_interaction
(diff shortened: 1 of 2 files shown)


app/models/link.rb
```diff
@@ -48,7 +49,34 @@ def self.build(params)
 
 private
 
-  def reset_time_and_status
-    self.update_columns(status: nil, link_last_checked: nil)
+  def find_link_url
```
davidbasalla: You could rename this to `link_with_matching_url`. I think that might be marginally clearer in terms of what the method returns. Same for `find_existing_link_url` and `find_existing_homepage_url`.
Also I think it's more standard to use nouns as Ruby method names when they return an object (Although I have no authoritative source to back that up :) )

app/models/link.rb
```diff
@@ -48,7 +49,34 @@ def self.build(params)
...
+    find_existing_link_url || find_existing_homepage_url
+  end
+
+  def find_existing_link_url
+    Link.where(url: self.url).distinct.first
```
davidbasalla: You could memoize this method and `find_existing_homepage_url`. It might save you some extra assignments later on in the code.

spec/models/link_spec.rb
```diff
@@ -111,13 +111,42 @@
...
+      @link = FactoryGirl.create(:link, url: "http://example.com/thing", status: "200", link_last_checked: Time.now)
+      @link.url = "http://www.angus.gov.uk"
+      @link.save!
+
+      expect(@link.status).to eq(@local_authority.status)
+      expect(@link.link_last_checked).to eq(@local_authority.link_last_checked)
+    end
+  end
+
+  describe "before_create" do
```
davidbasalla, line 144: We should add another test here for what happens when matching link can be found

emmabeynon:
Thanks @davidbasalla - we've updated based on your comments.

brenetic:
💥

PR #63, opened 2016-09-12
Make the link_status CSVs be generated from an exporter

klssmith:
- It will still generate them on the fly, but it'll be a bit more
  consistent with the big LinksExporter by being LinksStatus in
  lib/exporters and not relying on a presenter.
- We still need the method in the controller to be able to use
  `send_data`.

Trello: https://trello.com/c/a2GItwki/484-look-at-export-csv-generation-code-3

(With @issyl0.)


h-lame:
This looks good, keeping things close together will help us in future if we have more of them to write.

PR #64, opened 2016-09-12
Make importers more similar

h-lame:
For: https://trello.com/c/RUa7mvps/483-look-at-importers-3

This may be best reviewed commit-by-commit as the notes explain the thinking and process we went through to get here.  It will probably also be best reviewed if you check out the code and look at it in-situ rather than the diff view which will be messy.  Hopefully the `Processor`, `Summariser`, and callback-based approach is still easy to follow through, but if it's not then there's no point in this refactoring and we can try something else.


brenetic:
Good stuff!

PR #65, opened 2016-09-16
Use correct variables in Summariser#import_summary

h-lame:
While building we flipped from `data_source_xxx` to `import_source_xxx`
as the names for the internal variables to record the name and count
values for the imported data.  We changed the initializer and increment
methods, but forgot to change the reference in the `#import_summary`
methods.  Happily, ruby will treat all instance variable references as
`nil`, so nothing breaks but we do get odd log output.

### DIFF
--- a/lib/local-links-manager/import/summariser.rb
+++ b/lib/local-links-manager/import/summariser.rb
@@ -49,7 +49,7 @@ module LocalLinksManager
 
       def summary
         "#{@name} complete\n"\
-        "#{@data_items_name}: #{@data_items_count}\n"\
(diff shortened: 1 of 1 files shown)


h-lame:
I should have spotted this while working on #64, but I only ran the imports to check they still worked, didn't check the log output.  Bad me.  Consider this another reminder that using `attr_[reader|writer|accessor]` and using the method instead of the instance variable is a good idea as a typo will result in a `NoMethodError` instead of a `nil`.

brenetic:
👍

brenetic (approved):
🚀

PR #66, opened 2016-09-19
Opt-in to use the new GDS API Adapters behaviour

issyl0:
- Trello: https://trello.com/c/B5z6Dq8q/495.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '~> 30.2'
+  gem 'gds-api-adapters', '36.0.0'
 end
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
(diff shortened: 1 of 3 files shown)


brenetic (changes requested):
Looks good!

Can we just add a little more to the commit/PR message please

brenetic (approved):

davidbasalla (commented):
First day back from hols here, so I may have the wrong end of the stick here 😁

I was expecting that we would have to update some of our code to catch the possible Exceptions coming from GdsApi adapters with this change. Looking at the code, it looks like we only use the mapit adapter from GdsApiAdapters. So wondering whether we should do any work [here](https://github.com/alphagov/local-links-manager/blob/master/lib/local-links-manager/import/local_authorities_importer.rb#L94-L96) to catch exceptions that could now be thrown. Afaik previously a 404 or 410 response would have just returned `nil` here but now a Exception would just make stop code execution entirely?

However... we seem to be [rescuing in the Processor class](https://github.com/alphagov/local-links-manager/blob/master/lib/local-links-manager/import/processor.rb#L34) anyway, so probably everything is fine and will work anyway... so no changes needed, just me talking out loud :)

issyl0:
Interesting. I'd have expected some test failures if other things needed changing, but given we're rescuing everything, that probably explains it. I'll work out what to explicitly rescue for 404s and 410s, if I can.

davidbasalla:
Had another look through with @emmabeynon and think this looks good as is!

PR #70, opened 2016-11-03
Upgrade to Rails 5

issyl0:
- Upgrade Rails to v5.0.0.1 and other gems with `bundle update`.
- Update some of the config files with `bundle exec rails app:update`.
- Fix deprecation warnings when running tests.
- Upgrade the linter and correct errors (lots of them fixed by specifying a Ruby version in `.rubocop.yml`.
- Make models inherit from ApplicationRecord, not ActiveRecord::Base, as per http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-models-now-inherit-from-applicationrecord-by-default.
- Make jenkins.sh run bundle exec rails instead of bundle exec rake and include `bundle exec db:environment:set RAILS_ENV=test` in this file to make Jenkins builds work.

This wasn't too bad, compared to the last Rails upgrade I did (3 to 4)!


config/application.rb
```diff
@@ -1,15 +1,6 @@
...
-# Pick the frameworks you want:
-require "active_model/railtie"
-require "active_job/railtie"
-require "active_record/railtie"
-require "action_controller/railtie"
-require "action_mailer/railtie"
-require "action_view/railtie"
-require "sprockets/railtie"
-# require "rails/test_unit/railtie"
+require 'rails/all'
```
whoojemaflip: this will be loading ActionCable - which we don't need. Probably best to stick with the long form `requires`

config/initializers/wrap_parameters.rb
```diff
@@ -5,10 +5,10 @@
...
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json] if respond_to?(:wrap_parameters)
+  wrap_parameters format: [:json]
 end
 
 # To enable root element in JSON for ActiveRecord objects.
 # ActiveSupport.on_load(:active_record) do
-#  self.include_root_in_json = true
+#   self.include_root_in_json = true
```
whoojemaflip, line 13: 👾

config/environments/production.rb
```diff
@@ -1,8 +1,4 @@
 Rails.application.configure do
-  config.logstasher.enabled = true
```
whoojemaflip, line 2: Was this config important?

config/environments/production.rb
```diff
@@ -76,7 +69,17 @@
   config.active_support.deprecation = :notify
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
-  # config.log_formatter = ::Logger::Formatter.new
+  config.log_formatter = ::Logger::Formatter.new
```
whoojemaflip: might this break Kibana integration?

whoojemaflip (changes requested):
A few little things...

config/initializers/wrap_parameters.rb
issyl0: I thought about fixing it upstream (it was auto-generated by the `app:update` task), but found https://github.com/rails/rails/blob/master/CONTRIBUTING.md#did-you-fix-whitespace-format-code-or-make-a-purely-cosmetic-patch. 😞

issyl0: Hmm but actually, that diff is correct. That line is commented out, and before there was only one space after the #, not two as per convention.

issyl0: Hmm, actually, that diff is correct. It's an indent from the `ActiveSupport` block which - uncommented - would be fine, but as it's commented, it needs another space for consistency.

config/environments/production.rb
issyl0: The Publishing API's Rails upgrade diff leaves this commented out. I've just done the same.

config/environments/production.rb
issyl0: According to the Publishing API's Rails upgrade diff, yes. I've added it back.

config/application.rb
issyl0: Reverted and removed `config/cable.yml`.

whoojemaflip (approved):

config/environments/production.rb
whoojemaflip: and the following two lines please :)

whoojemaflip (approved):

PR #71, opened 2016-11-07
All councils page.

whoojemaflip:
A sortable, filterable index page to browse all local authorities. 

https://trello.com/c/6xQryalX/537-update-view-all-councils-page-5

llm-all-councils-implementation

Missing from the design is the fancy-pants search/filtering widget. There was some feedback from the user tests that I think we wanted to discuss before investing too much time in a build.  This solution uses the Table Filter from `govuk_admin_template` which does the basics for us.  A more sophisticated thing could be built on top of this, once we're sure what to go for.

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -19,6 +19,11 @@ class Link < ApplicationRecord
       .where(service_interactions: { service_id: service })
   }
 
+  HTTP_OK_STATUS_CODE = 200
+
+  scope :currently_broken, -> { where.not(status: HTTP_OK_STATUS_CODE) }
+  scope :have_been_checked, -> { where.not(status: nil) }
+
   def self.enabled_links
     self.joins(:service).where(services: { enabled: true })
   end
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,5 +1,6 @@
 class Link < ApplicationRecord
   before_update :set_time_and_status_on_updated_link, if: :url_changed?
+  after_save :trigger_recount_of_broken_links, if: :status_changed?
   before_create :set_time_and_status_on_new_link
 
   belongs_to :local_authority
@@ -76,5 +77,9 @@ private
   def set_status_and_last_checked_for(link)
     self.status = link.status
     self.link_last_checked = link.link_last_checked
+  end
+
+  def trigger_recount_of_broken_links
+    local_authority.calculate_count_of_broken_links
   end
 end
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,6 +1,5 @@
 class Link < ApplicationRecord
   before_update :set_time_and_status_on_updated_link, if: :url_changed?
-  after_save :trigger_recount_of_broken_links, if: :status_changed?
   before_create :set_time_and_status_on_new_link
 
   belongs_to :local_authority
@@ -82,9 +81,5 @@ private
   def set_status_and_last_checked_for(link)
     self.status = link.status
     self.link_last_checked = link.link_last_checked
-  end
-
-  def trigger_recount_of_broken_links
-    local_authority.calculate_count_of_broken_links
   end
 end
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -14,11 +14,11 @@ class LocalAuthority < ApplicationRecord
     Service.for_tier(self.tier).enabled
   end
 
-  def calculate_count_of_broken_links
-    # to ensure that we can replay our migrations, wrap the update in a guard clause.
-    if respond_to?(:broken_link_count)
-      update_attribute(:broken_link_count, self.links.where.not(status: 200).count)
-    end
+  def update_broken_link_count
+    update_attribute(
+      :broken_link_count,
+      links.have_been_checked.currently_broken.count
+    )
   end
 
 private
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -14,6 +14,13 @@ class LocalAuthority < ApplicationRecord
     Service.for_tier(self.tier).enabled
   end
 
+  def calculate_count_of_broken_links
+    # to ensure that we can replay our migrations, wrap the update in a guard clause.
+    if respond_to?(:broken_link_count)
+      update_attribute(:broken_link_count, self.links.where.not(status: 200).count)
+    end
+  end
+
 private
 
   def reset_time_and_status
--- a/app/views/local_authorities/index.html.erb
+++ b/app/views/local_authorities/index.html.erb
@@ -9,12 +9,15 @@
 <%= render partial: 'shared/home_nav' %>
 
 <% if @authorities.any? %>
-  <table class="table table-bordered table-with-counts" data-module="filterable-table">
+  <table class="table table-bordered table-with-counts table-with-controls" data-module="filterable-table">
     <thead>
       <tr class="table-header">
         <th>
           <div class='sort-order-dropdown'>
             <%= render partial: "shared/sort_order_dropdown" %>
+          </div>
+          <div class='filter-control'>
+            <%= render partial: "govuk_admin_template/table_filter_form" %>
           </div>
           <div class='count'>
             <span><%= @authorities.count %></span>
@@ -27,7 +30,7 @@
       <% @authorities.each do |authority| %>
         <tr>
           <td>
-            <%= link_to authority.name, local_authority_services_path(authority.slug), class: 'js-open-on-submit local-authority-name' %>
+            <%= link_to authority.name, local_authority_services_path(authority.slug), class: 'js-open-on-submit name' %>
             <div class="count">
               <span><%= authority.broken_link_count %></span>
               <%= 'Broken link'.pluralize(authority.broken_link_count) %>
--- a/app/views/local_authorities/index.html.erb
+++ b/app/views/local_authorities/index.html.erb
@@ -10,7 +10,12 @@
   <table class="table table-striped table-hover table-bordered contacts-table" data-module="filterable-table">
     <thead>
       <tr class="table-header">
-        <th>Local Authorities (<%= @authorities.count %>)</th>
+        <th>
+          <div class='count'>
+            <span><%= @authorities.count %></span>
+            <%= 'local authorities'.pluralize(@authorities.count) %>
+          </div>
+        </th>
       </tr>
       <%= render partial: "govuk_admin_template/table_filter",
       locals: {placeholder: "Filter list of local authorities"} %>
--- a/app/views/local_authorities/index.html.erb
+++ b/app/views/local_authorities/index.html.erb
@@ -9,7 +9,7 @@
 <%= render partial: 'shared/home_nav' %>
 
 <% if @authorities.any? %>
-  <table class="table table-striped table-hover table-bordered contacts-table" data-module="filterable-table">
+  <table class="table table-bordered table-with-counts" data-module="filterable-table">
     <thead>
       <tr class="table-header">
         <th>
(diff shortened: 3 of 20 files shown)


db/migrate/20161104150651_add_broken_link_count_to_local_authority.rb
```diff
@@ -0,0 +1,11 @@
+class AddBrokenLinkCountToLocalAuthority < ActiveRecord::Migration[5.0]
+  def up
+    add_column :local_authorities, :broken_link_count, :integer, default: 0
+
+    LocalAuthority.all.map(&:calculate_count_of_broken_links)
```
davidbasalla, line 5: Quick question, is this doing the same as `LocalAuthority.all.each { |la| la.calculate_count_of_broken_links }`? Does using the `map` method have any advantages here?

whoojemaflip: It's exactly the same functionally, but I feel the shorter syntax is a little more intention revealing. It can be a bit confusing the first time you see it though, I agree. Would you rather it were swapped out for the long-hand?

app/models/local_authority.rb
```diff
@@ -14,6 +14,13 @@ def provided_services
     Service.for_tier(self.tier).enabled
   end
 
+  def calculate_count_of_broken_links
+    # to ensure that we can replay our migrations, wrap the update in a guard clause.
+    if respond_to?(:broken_link_count)
+      update_attribute(:broken_link_count, self.links.where.not(status: 200).count)
```
davidbasalla: We were wondering if broken links == links.where.not(status: 200) and if there were any other status codes that might count as 'broken', such as redirects. But looking at the link checker it looks like redirects do return a 200 (if final redirects return a 200) so this looks fine. Briefly thought about defining a `broken` scope on the Link model to move the definition of what broken links are to the Link Model... not sure if necessary at this point

spec/models/link_spec.rb
```diff
@@ -155,4 +155,13 @@
     expect(@link.status).to be nil
     expect(@link.link_last_checked).to be nil
   end
+
+  describe "after_update" do
+    it "triggers a recount of the local authority's broken links" do
```
davidbasalla: Would it be worth adding another test here to demonstrate that `calculate_count_of_broken_links` will not get called if `status` does not change?

spec/models/local_authority_spec.rb
```diff
@@ -76,5 +76,13 @@
         expect(@local_authority.link_last_checked).to be_nil
       end
     end
+
+    describe "calculate_count_of_broken_links" do
```
davidbasalla: Small style niggle, I'm used to seeing method tests have a `#` in front of them, see http://betterspecs.org/#describe

davidbasalla: Also the word 'links' is missing from the test description

davidbasalla: We could add another link here with status 200 to show that the broken link count is correct in counting only links that are broken.

app/models/local_authority.rb
whoojemaflip: I really like that. It makes complete sense that when we talk about Broken Links that they be either a subset of all links aka. `Link.all.broken` as you suggest or even a first class model object eg. `class BrokenLink < Link`

spec/models/link_spec.rb
whoojemaflip: 👍

spec/models/local_authority_spec.rb
whoojemaflip: well spotted. #willfix

davidbasalla: Were also thinking that we wouldn't need to define a local authority as you could use the `local_authority` association from the link factory, but maybe it's more readable? _shrug_

db/migrate/20161104150651_add_broken_link_count_to_local_authority.rb
davidbasalla: No it's good like this, was just not used to seeing it.

app/views/local_authorities/index.html.erb
```diff
@@ -10,7 +10,9 @@
   <table class="table table-striped table-hover table-bordered contacts-table" data-module="filterable-table">
     <thead>
       <tr class="table-header">
-        <th>Local Authorities (<%= @authorities.count %>)</th>
+        <th>
+          <div class='count'><span><%= @authorities.count %></span> local authorities</div>
```
davidbasalla: Why did you add a "span" element here? Is that good practice generally?

whoojemaflip: I think github ate your markup.. which element are you referring to?

davidbasalla: ^ thanks, fixed :) 

So there's a spec which tests the wording that this commit changes. It gets updated in [another commit in this PR](https://github.com/alphagov/local-links-manager/pull/71/commits/c40c642b7615bee8347b3e93521bc662dd37536b). Might be worth either pulling that test change into this commit, or rolling this commit into the other one...

spec/support/appear_before_matcher.rb
```diff
@@ -0,0 +1,32 @@
+module SortOrderMatchers
```
davidbasalla, line 1: This is cool! Did you get this pattern from somewhere or is it completely hand rolled? I vaguely remember testing sort order can be a pain...

spec/factories/links.rb
```diff
@@ -5,6 +5,10 @@
     url { "http://#{local_authority.slug}.example.com/#{service_interaction.service.slug}/#{service_interaction.interaction.label.parameterize}" }
     status nil
     link_last_checked nil
+
+    trait :with_service_interaction do
```
davidbasalla: This trait is probably not needed because we've already defined the service_interaction association for this factory. We tried removing this trait and any calls to it and the tests still passed.

spec/models/local_authority_spec.rb
```diff
@@ -76,5 +76,13 @@
...
+      it "calculates the correct number of broken for the local authority" do
+        local_authority = FactoryGirl.create(:local_authority)
+        FactoryGirl.create(:link, :with_service_interaction, local_authority: local_authority, status: 500)
+        expect { local_authority.calculate_count_of_broken_links }.to change { local_authority.broken_link_count }.from(0).to(1)
```
davidbasalla: We can't actually get this test to pass. It appears that when the link instance is being created by FactoryGirl, the after_save callback is already being called so when we get to L84 the broken_link_count is already 1, not 0 as currently expected...

app/views/local_authorities/index.html.erb
```diff
@@ -19,14 +19,13 @@
...
-      locals: {placeholder: "Filter list of local authorities"} %>
     </thead>
     <tbody>
       <% @authorities.each do |authority| %>
         <tr>
-          <td class="name">
-            <%= link_to authority.name, local_authority_services_path(authority.slug), class: 'js-open-on-submit' %>
+          <td>
+            <%= link_to authority.name, local_authority_services_path(authority.slug), class: 'js-open-on-submit local-authority-name' %>
+            <div class="count"><span><%= authority.broken_link_count %> </span>Broken links</div>
```
davidbasalla: Could we add a test for this to the integration specs? Not sure how fiddly testing the rows would be...

davidbasalla:
This looks good, we've finished reviewing the commits and left notes accordingly. I didn't look at the CSS too much as it's really my 'fortay' but I'm sure it's good! :)

app/views/local_authorities/index.html.erb
whoojemaflip: the `span` element here gives us the ability to target the numeric bit of the sentence for fancy CSS formatting.

spec/support/appear_before_matcher.rb
whoojemaflip: I went looking for it on the internet, and rolled my own based on someone else's code which did 200% of what we needed here. Custom Matchers are the best.

whoojemaflip:
@davidbasalla @emmabeynon That was a thorough review! Thanks for pushing me to do better :)

I've rebased the suggested changes into the relevant commits to preserve the history.

As discussed, there's some light refactoring done to the `LinkStatusUpdater` class which allows us to simplify the `Link` and `LocalAuthority` classes from the initial design. Due to a race-hazard, it was a bunch more efficient to drop that refactoring into this PR rather than create several more blocker dependencies.

lib/local-links-manager/check_links/link_status_updater.rb
```diff
@@ -3,29 +3,27 @@
 module LocalLinksManager
   module CheckLinks
     class LinkStatusUpdater
```
emmabeynon, line 5: This looks good.  We should probably change similar code in the HomepageStatusUpdater class too so that it's consistent.

emmabeynon: Small change should be made to the commit message so it starts with an active verb.

app/models/link.rb
```diff
@@ -19,6 +19,11 @@ class Link < ApplicationRecord
       .where(service_interactions: { service_id: service })
   }
 
+  HTTP_OK_STATUS_CODE = 200
+
+  scope :currently_broken, -> { where.not(status: HTTP_OK_STATUS_CODE) }
```
emmabeynon, line 23: Would it be worth having tests for these new scopes?

app/models/link.rb
```diff
@@ -1,6 +1,5 @@
 class Link < ApplicationRecord
   before_update :set_time_and_status_on_updated_link, if: :url_changed?
-  after_save :trigger_recount_of_broken_links, if: :status_changed?
```
emmabeynon: Something we just thought of is that when an individual link is updated, without this `after_save` callback the broken link recount will not occur, and therefore the broken link count will be incorrect.  We should add some tests for this and figure out how to best trigger this recount for individually edited links.

app/models/link.rb
whoojemaflip: The code is run by the integration tests, so it seemed like overkill to write slow database-bound unit tests for something like this.

Additionally, unit testing a simple scope like this seems a bit like writing tests for ActiveRecord.  What do you think? I'm comfortable with the current level of coverage for this.

davidbasalla: Sounds good, agreed that the scope is fairly simple so not necessary to test further

lib/local-links-manager/check_links/link_status_updater.rb
```diff
@@ -8,15 +8,29 @@ def initialize(link_checker = LinkChecker.new)
       end
 
       def update
+        check_all_links
+        update_local_authorities_with_broken_link_count
```
davidbasalla: Just thinking about this... currently the broken link counts for LAs only get updated after all links have been checked. Currently this is run at 2am every night and takes about 6.5 hours to finish (estimate based on highly unscientific guesswork), so the broken link counts would only be updated around 8:30 am. Are we happy with this sort of delay in the results? Anyone using the service before 8:30 am would see wrong results...

lib/local-links-manager/check_links/link_status_updater.rb
whoojemaflip: I've just looked at the HomepageStatusUpdater class for the first time. It looks like the updater was maybe planned to be abstracted into common library code at some point, but that hadn't happened yet. I'm not sure what the benefits of consistency are, but I agree that it will need a look at some point.

What with the plans to build a more encompassing Link Checker, do you agree that we should leave this code as-is, for now at least?

app/models/link.rb
whoojemaflip: good catch

lib/local-links-manager/check_links/link_status_updater.rb
whoojemaflip: maybe we should recalculate after every link check completes?

lib/local-links-manager/check_links/link_status_updater.rb
brenetic: The one con of that is that the link checker is months away from discovery let alone actually building it. It will come after Lolotron V2 and mid (maybe after, depending . . . ) migration

brenetic:
👍

PR #74, opened 2016-11-10
Add services page

brenetic:
[Trello card](https://trello.com/c/zR4glUyX/538-build-view-all-services-page-3)

This adds the v2 Services page. See commit messages for more detail.

screen shot 2016-11-10 at 17 19 29

### DIFF
--- a/app/controllers/local_authorities_controller.rb
+++ b/app/controllers/local_authorities_controller.rb
@@ -4,11 +4,16 @@ class LocalAuthoritiesController < ApplicationController
   end
 
   def edit
-    @authority = LocalAuthority.find_by(slug: params[:slug])
+    @authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+  end
+
+  def show
+    @authority = LocalAuthorityPresenter.new(LocalAuthority.find_by_slug!(params[:local_authority_slug]))
+    @services = @authority.provided_services.order(lgsl_code: :asc)
   end
 
   def update
-    @authority = LocalAuthority.find_by(slug: params[:slug])
+    @authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
     @authority.homepage_url = params[:local_authority][:homepage_url].strip
     validate_and_save(@authority)
   end
@@ -18,7 +23,7 @@ private
   def validate_and_save(authority)
     if authority.save
       flash[:success] = "Homepage link has been saved."
-      redirect_to local_authority_services_path(local_authority_slug: params[:slug])
+      redirect_to local_authority_path(local_authority_slug: params[:local_authority_slug])
     else
       flash.now[:danger] = "Please enter a valid link."
       render :edit
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -11,7 +11,7 @@
 
 <% content_for :content do %>
 
-  <%= breadcrumbs style: :bootstrap %>
+  <%= breadcrumbs style: :bootstrap, display_single_fragment: true %>
 
   <%= yield %>
 <% end %>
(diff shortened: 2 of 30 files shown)


davidbasalla:
Looks like this needs to be rebased

app/controllers/local_authorities_controller.rb
```diff
@@ -6,11 +6,16 @@ def index
   end
 
   def edit
-    @authority = LocalAuthority.find_by(slug: params[:slug])
+    @authority = LocalAuthority.find_by(slug: params[:local_authority_slug])
+  end
+
+  def show
+    @authority = LocalAuthorityPresenter.new(LocalAuthority.find_by_slug!(params[:local_authority_slug]))
```
davidbasalla, line 11: Not necessary to fix now, but in the `update` method we use `LocalAuthority.find_by(slug: params[:local_authority_slug])` to get the local authority, so slightly differently from the `show` method. We could probably use the same method...

spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -8,4 +8,17 @@
...
+
+  describe "GET #show" do
+    before do
+      @local_authority = FactoryGirl.create(:local_authority, name: 'Angus')
+      @service = FactoryGirl.create(:service, label: 'Service 1', lgsl_code: 1)
+    end
+
+    it "returns http success" do
+      login_as_stub_user
+      get :index, params: { local_authority_slug: @local_authority.slug, service_slug: @service.slug }
```
davidbasalla: Should this be `get :show`?

app/views/layouts/application.html.erb
```diff
@@ -11,7 +11,7 @@
 
 <% content_for :content do %>
 
-  <%= breadcrumbs style: :bootstrap %>
+  <%= breadcrumbs style: :bootstrap, display_single_fragment: true %>
```
issyl0, line 14: Does this mean that the breadcrumb shows up on the first page too?

spec/features/services/service_index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  it "displays a filter box" do
+    expect(page).to have_selector('.filter-control-full-width')
+  end
+
+  it "shows services sorted by broken link count" do
+    expect(page).to have_content('2 services')
+    expect(page).to have_content('99 Zebra Fouling 1')
+    expect(page).to have_content('1 Aardvark Wardens 0')
+    expect('Zebra Fouling').to appear_before('Aardvark Wardens')
```
issyl0, line 25: 😂

spec/controllers/services_controller_spec.rb
```diff
@@ -1,4 +1,11 @@
 require 'rails_helper'
 
 RSpec.describe ServicesController, type: :controller do
```
davidbasalla, line 3: As an aside, I think we briefly talked about this, but should we switch to using requests specs going forward rather than controller tests (as that seems to be the Rails 5 way)? Would be interesting to have a go at converting the old specs at some point too.

app/views/layouts/application.html.erb
brenetic: Yes :)

app/controllers/local_authorities_controller.rb
brenetic: 🌮

app/views/layouts/application.html.erb
issyl0: I'm not sure about this - it seems weird to me - you're already on the page and there's nothing to go back to, so why not only display breadcrumbs when there's a trail? But it's in the prototype, so let's go with it.

PR #75, opened 2016-11-11
Update view all councils page

whoojemaflip:
Implements product review feedback from [this trello card](https://trello.com/c/zwuEYa60/548-update-to-view-all-councils-page-following-product-review)


config/locales/en.yml
```diff
@@ -1,24 +1,2 @@
...
-#
-# This would use the information in config/locales/es.yml.
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
-
 en:
-  broken_link_count: Number of broken links
-  name: A-Z
+  hello: hello
```
issyl0, line 2: I know this was already there in the first place, but do we need it (given we've deleted the comments)? I suppose it serves as a good example for what to do in this file, if people don't already know.

whoojemaflip: I think that YAML requires both a key and a value to be valid. We could always delete the file, but that seems like overkill. wdyt?

issyl0: Ah right. Leave it then. 👍

PR #76, opened 2016-11-11
Add Specific Service Page

whoojemaflip:
Implements the Specific Service Page. See commits for more specific context.

[Trello card](https://trello.com/c/AfsQXUa3/539-build-specific-service-page-5)

screen shot 2016-11-18 at 11 00 14

Mobbed by @whoojemaflip @deborahchua @brenetic


davidbasalla:
Sorry in advance for the pedantry, but prefer to have commit messages in the format of Active verb in present tense first - eg https://github.com/alphagov/styleguides/blob/master/git.md#structure

davidbasalla:
The [first commit](https://github.com/alphagov/local-links-manager/pull/76/commits/04eceb54ee3db0267e3dd8011d0b05283913f9f3) is a bit confusing in that adds specs changes which get removed in a later commit. It also refers to the old style tiers, so I suspect that commit was made before the Tier model refactoring was done. Also it does not include the switch over of the index page to show page. So maybe some rebasing/squashing is required?

brenetic:
@davidbasalla Done and done 🚀

app/presenters/any_link_presenter.rb
```diff
@@ -0,0 +1,92 @@
+class AnyLinkPresenter
```
davidbasalla: Would it be worth adding unit tests for this class? Also we still have the `LinkPresenter` and `LocalAuthorityPresenter` which appear to do similar things. Is there a possibility of getting rid of any of these classes?

app/presenters/any_link_presenter.rb
```diff
@@ -0,0 +1,92 @@
...
+    presenter_klass.new(link_container, *args)
+  end
+
+  class ForALink < SimpleDelegator
+    include UrlStatusPresentation
+    attr_reader :with_service_links_count
+
+    def initialize(link, with_service: nil, with_service_links_count: 0, context:, local_authority: nil)
+      super(link)
+      @with_service = with_service
```
davidbasalla: Totally minor, but could the `with_` be removed or does it add any sort of special meaning here? Both for `with_service` and `with_service_links_count`

app/presenters/any_link_presenter.rb
```diff
@@ -0,0 +1,92 @@
...
+      @with_service_links_count = with_service_links_count
+      @context = context
```
davidbasalla: Is there a better name for `context`? It's not immediately clear to me what that refers to.

app/presenters/any_link_presenter.rb
```diff
@@ -0,0 +1,92 @@
+class AnyLinkPresenter
+  def self.for(link_container, *args)
+    presenter_klass =
+      case link_container
+      when Link
+        ForALink
+      when LocalAuthority
```
davidbasalla: Does this code branch ever get called? It looks like we only use the `AnyLinkPresenter` class on the services show page for Links, but never for LocalAuthorities? If so, then we should remove this, and maybe revisit what this class does as opposed to the already existing LinkPresenter?

app/presenters/any_link_presenter.rb
h-lame: `context` is an object that we can use in the presenter to call action view helper methods (in this case we use it to call the named route helpers).  It's usually a rails controller, or the action view template where the presenter is constructed.   Perhaps `view_context` or `template` would be a better name?

app/views/services/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <% @local_authorities.each do |local_authority| %>
+        <% links_by_authority = @links[local_authority.id] %>
+        <% next if links_by_authority.blank? %>
+        <% presented_authority = LocalAuthorityPresenter.new(local_authority) %>
+        <tr data-local-authority-id="<%= presented_authority.id %>" class="header-row">
+          <th class="name" colspan="4">
+            <h3><%= link_to presented_authority.name, local_authority_path(presented_authority.slug) %></h3>
+          </th>
+        </tr>
+        <% first_link, *remaining_links = *links_by_authority %>
```
davidbasalla: Could you explain what this line does? Not used to seeing `*remaining_links = *links_by_authority` style notation

app/presenters/any_link_presenter.rb
h-lame: This whole presenter may have been brought over from the prototype prematurely.  In the prototype we wanted to present all links, including the home page link in the same list view, so it was useful to have a single object that would act appropriately given a `Link` or a `LocalAuthority`.  If we're not doing that anymore it might not be useful, and you could put the extra methods from here into `LinkPresenter`.

app/views/services/show.html.erb
h-lame: This could just be `first_link, *remaining_links = links_by_authority` as the splat on `links_by_authority` is redundant.  The splat on `remaining_links` means to collect all the rest of the values.  E.g.:

``` ruby
first, rest = [1,2,3,4,5]
first #=> 1
rest #=> 2

first, *rest = [1,2,3,4,5]
first #=> 1
rest #=> [2,3,4,5]
```

brenetic: Takes the first item from `links_by_authority` and returns it in `first_link`. Any remaining links are returned in `remaining_links`. @h-lame magic

whoojemaflip: one of the finest bits of functional ruby I've seen in a long while. A++++ would copy Murray's prototype code again.

app/presenters/any_link_presenter.rb
whoojemaflip: I'm not convinced that unit-tests will add much value here. The code is getting exercised in the feature specs, which is where we test the rest of the view layer.  Open to counter-arguments of course.

app/presenters/any_link_presenter.rb
whoojemaflip: renamed to `view_context` for clarity.

brenetic:
@davidbasalla @issyl0 @emmabeynon This PR has been updated since the last comments. Ready for another viewing 🚀

davidbasalla:
Looks good! Wondering whether we can retire the `LinkPresenter` class, but it's still used for Interactions at the moment, so we can probably do this at a later point.

PR #77, opened 2016-11-14
[DO NOT MERGE] Remodelling Tier relationships

whoojemaflip:
The concept of `Tiers` was modelling as text strings with an opaque relationship. 
e.g. a Local Authority might be a 'unitary' type, and a Service might be of type 'district/unitary'
These implicit joins were slow and hard to reason about.

This replaces that with DB relations so we can enjoy all the fun stuff that brings.

### DIFF
--- a/db/migrate/20161114122001_create_tier_and_service_tier.rb
+++ b/db/migrate/20161114122001_create_tier_and_service_tier.rb
@@ -1,0 +1,76 @@
+class CreateTierAndServiceTier < ActiveRecord::Migration[5.0]
+  def up
+    # create tables
+    create_table :tiers do |t|
+      t.string :name, null: false, unique: true
+      t.timestamps
+    end
+
+    create_table :service_tiers do |t|
+      t.integer :service_id
+      t.integer :tier_id
+      t.timestamps
+    end
+
+    add_column :local_authorities, :tier_id, :integer, index: true
+
+    # migrate data
+    unitary_tier = Tier.create(name: 'unitary')
+    district_tier = Tier.create(name: 'district')
+    county_tier = Tier.create(name: 'county')
+
+    LocalAuthority.all.each do |la|
+      la.tier_id = Tier.find_by(name: la.tier).id
+      la.save
+    end
+
+    Service.where.not(tier: nil).each do |service|
+      case service.tier
+      when 'district/unitary'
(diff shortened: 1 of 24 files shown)


db/migrate/20161114122001_create_tier_and_service_tier.rb
```diff
@@ -0,0 +1,76 @@
...
+        ServiceTier.create(service: service, tier: unitary_tier)
+      when 'all'
+        ServiceTier.create(service: service, tier: county_tier)
+        ServiceTier.create(service: service, tier: unitary_tier)
+        ServiceTier.create(service: service, tier: district_tier)
+      end
+    end
+
+    # delete old data
+    remove_column :local_authorities, :tier
```
whoojemaflip, line 43: We should hold off deleting the old data until we're confident that everything works

db/migrate/20161114122001_create_tier_and_service_tier.rb
```diff
@@ -0,0 +1,76 @@
...
+    unitary_tier = Tier.create(name: 'unitary')
+    district_tier = Tier.create(name: 'district')
+    county_tier = Tier.create(name: 'county')
+
+    LocalAuthority.all.each do |la|
+      la.tier_id = Tier.find_by(name: la.tier).id
+      la.save
+    end
+
+    Service.where.not(tier: nil).each do |service|
```
whoojemaflip, line 27: This code is duplicated here https://github.com/alphagov/local-links-manager/pull/77/files#diff-03421abb293f9a55667d48ebcfdc6f48R55 - should it be extracted into a common method?

db/migrate/20161114122001_create_tier_and_service_tier.rb
```diff
@@ -0,0 +1,76 @@
...
+    create_table :service_tiers do |t|
+      t.integer :service_id
+      t.integer :tier_id
+      t.timestamps
+    end
+
+    add_column :local_authorities, :tier_id, :integer, index: true
+
+    # migrate data
+    unitary_tier = Tier.create(name: 'unitary')
```
whoojemaflip, line 18: Could we structure this in such a way that the data-migration could be replayed? Leaning towards a rake task - is there a way of running these in production?

whoojemaflip:
This needed splitting into multiple PRs so as to be deployed.

PR #81, opened 2016-11-16
[3 of 4] Tiers - remove deprecated 'tier' columns

whoojemaflip:
Tidies up the schema cruft left over from previous work.

*NOTE:* This must not be deployed to any server until [this PR](https://github.com/alphagov/local-links-manager/pull/78) and [this PR](https://github.com/alphagov/local-links-manager/pull/79) have been released OK.

PR #79, opened 2016-11-16
[2 of 4] Adds new code paths for working with Tier relations

whoojemaflip:
Note: this can only be deployed once the [PR with migrations](https://github.com/alphagov/local-links-manager/pull/78) has been
deployed and [`migrate_tiers` rake task](https://github.com/alphagov/local-links-manager/blob/master/lib/tasks/migrate_tiers_to_new_schema.rake) run.

PR #82, opened 2016-11-16
[4 of 4] Adds the bi-directional relation to LocalAuthority and Service models

whoojemaflip:
Previous Pull Requests have created the relationship between `Local Authorities` and `Services` via `Tiers` and the `ServiceTiers` join table. [[3 of 4]](https://github.com/alphagov/local-links-manager/pull/81) deleted the `tier` columns from both of these tables, which frees up the namespace for this change.

In addition, it has been possible to create the sugar methods `LocalAuthority#tier => "district"` and `Service#tiers => ["district", "unitary"]`
 
**NOTE** This cannot be deployed until [[1]](https://github.com/alphagov/local-links-manager/pull/78), [[2]](https://github.com/alphagov/local-links-manager/pull/79), and [[3]](https://github.com/alphagov/local-links-manager/pull/81) have been released.

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -4,7 +4,7 @@ class LocalAuthority < ApplicationRecord
   validates :gss, :snac, :slug, uniqueness: true
   validates :gss, :name, :snac, :slug, presence: true
   validates :homepage_url, non_blank_url: true, allow_blank: true
-  validates :tier_id, allow_blank: true, inclusion:
+  validates :tier_id, presence: true, inclusion:
     {
       in: [Tier.unitary, Tier.district, Tier.county],
       message: "%{value} is not a valid tier"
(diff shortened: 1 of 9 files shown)


spec/lib/local-links-manager/import/local_authorities_importer_spec.rb
```diff
@@ -31,7 +31,7 @@ def fixture_file(file)
 
           expect(la.name).to eq("Aberdeen City Council")
           expect(la.snac).to eq("00QA")
-          expect(la.tier_id).to eq(Tier.unitary)
+          expect(la.tier).to eq('unitary')
```
davidbasalla, line 34: The changes in this file should probably be rolled into the last commit as that introduced the `tier` method for LocalAuthority

app/models/local_authority.rb
```diff
@@ -13,6 +13,10 @@ class LocalAuthority < ApplicationRecord
   has_many :links
   belongs_to :parent_local_authority, foreign_key: :parent_local_authority_id, class_name: "LocalAuthority"
 
+  def tier
```
davidbasalla, line 18: For the `tiers` method for Service model we added a unit test, but not for this similar method. Was that intentional as this method is not as complex? Might be a useful thing to document in the tests anyway?

davidbasalla:
Looks good, only 2 minor comments!

whoojemaflip:
@davidbasalla  Done! Rebased 🚀

davidbasalla:
Great! 👍

PR #84, opened 2016-11-21
Add Jenkinsfile

afda16:
We are currently looking at updating the way in which we do CI. In Jenkins 2 we can define our Build steps with a Jenkinsfile. This is one of the applications we would like to test with, because it's simple.

This Jenkinsfile is for Jenkins 2. The new version of Jenkins takes
care of notifying GitHub so we can remove `gh-status`.

There are probably other things in the `jenkins.sh` which can be deduplicated
as Jenkinsfile functions.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,34 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'local-links-manager'
+
+node {
+   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+   try {
+      stage("Checkout") {
+         checkout scm
+      }
+
+      stage("Build") {
+         sh "${WORKSPACE}/jenkins.sh"
+      }
+
+      stage("Push release tag") {
+         echo 'Pushing tag'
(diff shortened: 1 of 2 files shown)


brenetic:
Is this a WIP?

afda16:
This is ready to review and merge

Jenkinsfile
```diff
@@ -0,0 +1,34 @@
...
+         checkout scm
+      }
+
+      stage("Build") {
+         sh "${WORKSPACE}/jenkins.sh"
+      }
+
+      stage("Push release tag") {
+         echo 'Pushing tag'
+         //govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
```
brenetic: Should this remain commented out?

alexmuller: I think it can be uncommented in this repo because it's not enabled here yet: https://github.com/alphagov/govuk-puppet/blob/0575de6cf1f11e4f74ce06b1862eb3e72e7010eb/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L18

afda16:
I updated the Jenkinsfile with your last comment

PR #86, opened 2016-11-22
Make specific service page filterable

whoojemaflip:
The service page now shows the Broken links, Good links and All links tabs, as well as total link count sorted alphabetically by local authority name.

screen shot 2016-11-22 at 15 04 10

[Trello card](https://trello.com/c/eVRqki9r/551-make-specific-service-page-filterable-sortable-and-searchable-3)

Mobbed by @issyl0 @whoojemaflip @brenetic


davidbasalla:
It looks like the rendering of the templates is taking a long time, make the page loading a bit slow. I know we mentioned this before, is there anything obvious we could do the speed that up?

app/controllers/services_controller.rb
```diff
@@ -5,9 +5,30 @@ def index
...
+      .links
+      .where(link_filter_sql)
+      .includes([:service, :interaction, :local_authority])
+      .all
+  end
 
-    @links = @service.links.includes([:service, :interaction, :local_authority]).all.group_by(&:local_authority_id)
+  def link_filter_sql
+    case params[:filter]
+    when 'broken_links'
```
davidbasalla, line 26: Can we use the `currently_broken` scope in the Link model instead, as we've already defined there what a broken link is?  If so, we could also create a scope for 'good links'.

brenetic:
@davidbasalla if you touch `tmp/caching-dev.txt` on your dev vm caching will work and it will be faster

app/controllers/services_controller.rb
```diff
@@ -5,9 +5,30 @@ def index
...
+      "links.status != '200' OR links.status IS NULL"
+    when 'good_links'
+      "links.status = '200'"
+    else
+      "1 = 1"
```
davidbasalla: What does this do?

davidbasalla: What does this do?

spec/requests/api/link_spec.rb
```diff
@@ -107,12 +107,12 @@
...
-        interaction_2 = FactoryGirl.create(:interaction, label: 'appeal', lgil_code: 2)
-        service_interaction_1 = FactoryGirl.create(:service_interaction, service: service, interaction: interaction_1)
-        service_interaction_2 = FactoryGirl.create(:service_interaction, service: service, interaction: interaction_2)
-        FactoryGirl.create(:link, local_authority: local_authority, service_interaction: service_interaction_1)
-        FactoryGirl.create(:link, local_authority: local_authority, service_interaction: service_interaction_2)
+        interaction_1 = create(:interaction, label: 'report', lgil_code: 1)
+        interaction_2 = create(:interaction, label: 'appeal', lgil_code: 2)
+        service_interaction_1 = create(:service_interaction, service: service, interaction: interaction_1)
+        service_interaction_2 = create(:service_interaction, service: service, interaction: interaction_2)
+        create(:link, local_authority: local_authority, service_interaction: service_interaction_1, url: "http://blackburn.example.com/abandoned-shopping-trolleys/report")
```
davidbasalla, line 114: Just checking why we need to explicitly declare this url?

whoojemaflip: these tests were asserting on a specific URL that was being generated in the Link Factory. When we changed the factory class, these unrelated tests started failing, so we updated them in the simplest possible way. Next time anyone is in the codebase in this vicinity I'd recommend a rewrite TBH.

app/controllers/services_controller.rb
whoojemaflip: it's gross.  I like your suggestion better :)

app/controllers/services_controller.rb
```diff
@@ -14,21 +14,21 @@ def show
 private
 
   def links_for_service
-    @_links_for_service ||= @service
-      .links
-      .where(link_filter_sql)
+    @_links_for_service ||= filtered_links
```
davidbasalla, line 17: Nice! I reckon we could just squash this into the original commit rather than have it as a separate commit, so it's less confusing when we look back at the commit history

whoojemaflip: Good thinking. 🍏 🤔

emmabeynon:
LGTM 👍

PR #87, opened 2016-11-22
Configure Memcached as cache store

whoojemaflip:
In order to use the Rails fragment cache to speed up our render times,
we need some kind of cache-store. Other gov.uk Rails apps are using
memcached which is available to backend services without further
configuration, making it an easy choice.

As this change touched the Gemfile, it seemed sensible to leave it in a better state than it started.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,8 @@ gem 'redis-namespace', '1.5.2'
 
 gem 'whenever', require: false
 
+gem 'dalli'
+
 group :development, :test do
   gem 'pry-rails'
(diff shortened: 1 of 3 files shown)


brenetic:
🔤 ❗️ 👍  LGTM FFTM

sihugh:
The Gemfile is so very nearly alphabetic now...

whoojemaflip:
I missed a bit, didn't I. Coming up

whoojemaflip:
:eyes:

PR #88, opened 2016-11-23
Make specific council page filterable

whoojemaflip:
The Local Authority page now shows the Broken links, Good links and All links tabs, as well as total link count sorted alphabetically by service.

screen shot 2016-11-23 at 14 13 23

[Trello Card](https://trello.com/c/0UprwqoR/552-make-specific-council-page-filterable-sortable-and-searchable-2)

Mobbed by @whoojemaflip, @brenetic and @issyl0

### DIFF
--- a/app/controllers/local_authorities_controller.rb
+++ b/app/controllers/local_authorities_controller.rb
@@ -13,6 +13,7 @@ class LocalAuthoritiesController < ApplicationController
     @link_filter = params[:filter]
     @services = @authority.provided_services.order('services.label ASC')
     @links = links_for_authority.group_by { |link| link.service.id }
+    @link_count = links_for_authority.count
   end
 
   def update
--- a/app/controllers/local_authorities_controller.rb
+++ b/app/controllers/local_authorities_controller.rb
@@ -10,8 +10,9 @@ class LocalAuthoritiesController < ApplicationController
 
   def show
     @authority = LocalAuthorityPresenter.new(LocalAuthority.find_by_slug!(params[:local_authority_slug]))
+    @link_filter = params[:filter]
     @services = @authority.provided_services.order('services.label ASC')
-    @links = @authority.links.includes([:service, :interaction]).all.group_by { |link| link.service.id }
+    @links = links_for_authority.group_by { |link| link.service.id }
   end
 
   def update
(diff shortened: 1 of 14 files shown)


app/controllers/local_authorities_controller.rb
```diff
@@ -21,6 +22,26 @@ def update
...
 
+  def links_for_authority
+    @_links_for_authority ||= filtered_links
+      .includes([:service, :interaction])
+      .all
+  end
+
+  def filtered_links
+    # Returns the list of links for enabled services only.
+    links = @authority.links.joins(:service).merge(@authority.provided_services)
```
emmabeynon: Can we use the `enabled_links` method on the Link model here instead?

spec/features/local_authorities/local_authority_show_spec.rb
```diff
@@ -70,6 +70,16 @@
 
     let(:http_status) { 200 }
 
+    it 'shows a count of the number of all links for enabled services' do
+      within('thead') do
+        expect(page).to have_content "2 link"
```
emmabeynon: Just a small language thing - should this be "2 links"?

app/controllers/local_authorities_controller.rb
whoojemaflip: There's a subtle difference between the two, which the code comment doesn't address. Now you've pointed it out, I'm not sure which is the right solution.

For example:
```
> @authority.links.enabled_links.count
SELECT COUNT(*) 
FROM "links" 
INNER JOIN "service_interactions" ON "service_interactions"."id" = "links"."service_interaction_id" 
INNER JOIN "services" ON "services"."id" = "service_interactions"."service_id" 
WHERE "links"."local_authority_id" = $1 
AND "services"."enabled" = $2  
[["local_authority_id", 41], ["enabled", true]]
=> 111
```

vs 

```
> @authority.links.joins(:service).merge(@authority.provided_services).count
SELECT COUNT(*) 
FROM "links" 
INNER JOIN "service_interactions" ON "service_interactions"."id" = "links"."service_interaction_id" 
INNER JOIN "services" ON "services"."id" = "service_interactions"."service_id" 
INNER JOIN "service_tiers" ON "services"."id" = "service_tiers"."service_id" 
WHERE "links"."local_authority_id" = $1 
AND "service_tiers"."tier_id" = $2 
AND "services"."enabled" = $3  
[["local_authority_id", 41], ["tier_id", 2], ["enabled", true]]
=> 97
```

In this case, the latter example produces fewer results because it is only showing links for services of the appropriate Tier for the local authority.  It's interesting, as it shows we must have some out-of-date info in our database already.

issyl0:
This needs a rebase.

PR #89, opened 2016-11-24
Throw a real error in the no data case

whoojemaflip:
Tidies up some old code paths. 

[Trello Card](https://trello.com/c/Jlnm50HW/563-throw-a-real-error-in-the-no-data-case-in-local-links-manager-2)


emmabeynon:
LGTM 👍

PR #90, opened 2016-11-24
Update the specific council service page based on the prototype

deborahchua:
We've changed the routes to not be
`local-authorities/local_authority_name/services/service_name/interactions`
but instead `local-authorities/local_authority_name/interaction_name` as
per the prototype layout. Fixed all the failing tests as a result of this.

We also took out the LGIL code, and replaced "LGSL" with "Service code" as
a description. We also removed the homepage URL displaying.

Lastly, we removed the ability for people to add new links, pending further
user research.

Trello: https://trello.com/c/zHKGufLG/562-specific-council-and-service-page-2

Mobbed with @deborahchua, @brenetic and @issyl0.

### DIFF
--- a/app/views/interactions/index.html.erb
+++ b/app/views/interactions/index.html.erb
@@ -2,36 +2,30 @@
 
 <% breadcrumb :interactions, @authority, @service %>
 
-<%= render partial: "shared/local_authority_details", locals: {authority: @authority} %>
-
 <div class="page-title">
+  <h1><%= @authority.name %></h1>
+
   <h2><%= @service.label %></h2>
+  <p><b>Service code</b> <%= @service.lgsl_code %></p>
 </div>
 
 <% if flash[:success] %>
 <div class="alert alert-success"><%= flash[:success] %></div>
 <% end %>
 
-<p><b>LGSL</b> <%= @service.lgsl_code %></p>
-
 <% if @interactions.any? %>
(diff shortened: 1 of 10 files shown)


app/views/interactions/index.html.erb
```diff
@@ -2,36 +2,30 @@
...
 
 <% if flash[:success] %>
 <div class="alert alert-success"><%= flash[:success] %></div>
 <% end %>
 
-<p><b>LGSL</b> <%= @service.lgsl_code %></p>
-
 <% if @interactions.any? %>
-  <table class="table table-striped table-hover table-bordered contacts-table" data-module="filterable-table">
+  <table class="table table-striped table-hover table-bordered contacts-table">
```
whoojemaflip: Based on earlier feedback from Mark, I think the table should NOT include the `table-striped`, `table-hover` and `contacts-table` classes.

config/routes.rb
```diff
@@ -8,9 +8,8 @@
   resources 'services', only: [:index, :show], param: :service_slug
 
   scope '/local_authorities/:local_authority_slug/services/:service_slug' do
-    resources "interactions", only: [:index], param: :slug do
-      resource "links", only: [:edit, :update, :destroy]
-    end
+    root to: 'interactions#index', as: 'local_authority_with_service'
```
whoojemaflip: annoying question: would this be better expressed as a `links` resource?

whoojemaflip:
:+1:

PR #91, opened 2016-11-24
Update edit link page

brenetic:
- The edit link page now has both the service title and the interaction
title
- 'LGSL code' is now 'Service code'
- Removed LGIL
- 'URL' is now 'Page URL'
- Removed the ability to edit the homepage url
- 'Save' is now 'Update'
- Confirmation before deletion message
- Return to correct page upon update, cancel and delete of a link

# Before

before

# After

after

screen shot 2016-11-25 at 12 34 47

[Trello card](https://trello.com/c/gh6xcdeY/564-update-edit-links-and-add-link-pages-2)

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -3,6 +3,39 @@ require 'local-links-manager/export/link_status_exporter'
 class LinksController < ApplicationController
   before_action :load_dependencies
 
+  def edit
+    flash[:back] = request.env['HTTP_REFERER']
+    if flash[:link_url]
+      @link.url = flash[:link_url]
+      @link.validate
+    end
+  end
+
+  def update
+    @link.url = link_url
+
+    if @link.save
+      @link.local_authority.update_broken_link_count
+      @link.service.update_broken_link_count
+      redirect
+    else
+      flash[:danger] = "Please enter a valid link."
+      flash[:back] = flash[:back]
+      redirect_back
+    end
+  end
+
+  def destroy
+    if @link.destroy
+      redirect('deleted')
+    else
+      flash.now[:danger] = "Could not delete link."
+      flash[:back] = flash[:back]
+      flash[:danger] = "Could not delete link."
+      redirect_back
+    end
+  end
+
   def homepage_links_status_csv
     data = LocalLinksManager::Export::LinkStatusExporter.homepage_links_status_csv
     send_data data, filename: "homepage_links_status.csv"
@@ -13,30 +46,6 @@ class LinksController < ApplicationController
     send_data data, filename: "links_status.csv"
   end
 
-  def edit; end
-
-  def update
-    @link.url = params[:link][:url].strip
-
-    if @link.save
-      @link.local_authority.update_broken_link_count
-      @link.service.update_broken_link_count
-      redirect
-    else
-      flash.now[:danger] = "Please enter a valid link."
-      render :edit
-    end
-  end
-
-  def destroy
-    if @link.destroy
-      redirect('deleted')
-    else
-      flash.now[:danger] = "Could not delete link."
-      render :edit
-    end
-  end
-
 private
 
   def load_dependencies
@@ -44,6 +53,19 @@ private
     @interaction = Interaction.find_by(slug: params[:interaction_slug])
     @service = Service.find_by(slug: params[:service_slug])
     @link = Link.retrieve(params)
+  end
+
+  def set_back_url_before_post_request
+    flash[:back] = flash[:back] || request.env['HTTP_REFERER']
+  end
+
+  def link_url
+    params[:link][:url].strip
+  end
+
+  def redirect_back
+    flash[:link_url] = link_url
+    redirect_to edit_link_path(@local_authority, @service, @interaction)
   end
 
   def redirect(action = 'saved')
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -2,9 +2,10 @@ require 'local-links-manager/export/link_status_exporter'
 
 class LinksController < ApplicationController
   before_action :load_dependencies
+  before_action :set_back_url_before_post_request, only: [:edit, :update, :destroy]
+  helper_method :back_url
 
   def edit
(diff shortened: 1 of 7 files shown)


brenetic:
The commit for e98b2fbfefdd51ca48b9bd2190cd66b8665ad70e will be removed once https://github.com/alphagov/local-links-manager/pull/91 has been merged

app/views/links/edit.html.erb
```diff
@@ -1,30 +1,32 @@
...
+<h3><%= @interaction.label %></h3>
 
 <%= form_for @link, url: link_path(@local_authority.slug, @service.slug, @interaction.slug), method: "put" do |f| %>
   <div class="form-group <% if flash[:danger] %>has-error<% end %>">
     <div class='col-xs-12 no-gutter'>
-      <%= f.label :url, 'URL' %>
+      <%= f.label :url, 'Page URL' %>
 
-      <p class="text-muted">eg http://www.local-authority.gov.uk</p>
+      <p class="text-muted">example http://www.local-authority.gov.uk</p>
```
issyl0: ISTR this should be "for example", otherwise LGTM pending the PR it relies upon being merged.

app/controllers/links_controller.rb
```diff
@@ -46,9 +51,20 @@ def load_dependencies
...
+  def set_back_url_before_post_request
+    flash[:back] = flash[:back] || request.env['HTTP_REFERER']
+  end
+
+  def redirect_on_failed_validation
+    flash[:link_url] = params[:link][:url].strip
+    redirect_to edit_link_path(@local_authority, @service, @interaction)
+  end
+
   def redirect(action = 'saved')
```
whoojemaflip, line 78: Could this be called `redirect_back` to give a clearer idea of what it's doing?

app/views/interactions/index.html.erb
```diff
@@ -25,7 +25,12 @@
     <tbody>
       <% @interactions.each do |interaction| %>
         <% next if interaction.link_url.nil? %>
-        <tr id="<%= interaction.lgil_code %>" <% if flash[:lgil] == interaction.lgil_code %>class="success"<% end %>>
+        <tr id="<%= interaction.lgil_code %>"
+        <% if flash[:updated].present? &&
```
whoojemaflip: if only there were some kind of Presenter to delegate this logic to...

app/views/links/edit.html.erb
```diff
@@ -25,7 +25,7 @@
     </div>
 
     <div class='actions add-vertical-margins col-xs-6 no-gutter'>
-      <%= link_to 'Cancel', local_authority_with_service_path(local_authority_slug: @local_authority.slug, service_slug: @service.slug), class: 'btn btn-default add-right-margin' %>
+      <%= link_to 'Cancel', flash[:back], class: 'btn btn-default add-right-margin' %>
```
whoojemaflip: instead of using `flash[:back]` here, would it be possible to use a helper method to wrap this functionality, and provide a fallback?

app/controllers/links_controller.rb
```diff
@@ -46,6 +54,11 @@ def load_dependencies
     @link = Link.retrieve(params)
   end
 
+  def redirect_on_failed_validation
+    flash[:link_url] = params[:link][:url].strip
```
whoojemaflip: There's some duplication of `params[:link][:url].strip` - could this be extracted into a method?

app/presenters/url_status_presentation.rb
```diff
@@ -23,4 +23,10 @@ def last_checked
       "Link not checked"
     end
   end
+
+  def updated_url?
```
whoojemaflip: It might be more idiomatic to call this `updated?`

spec/features/local_authorities/local_authority_show_spec.rb
```diff
@@ -97,6 +97,24 @@
...
     end
 
+    context 'editing a link' do
+      it 'returns you to the correct page after updating a link' do
+        within('.table') { click_on('Edit link', match: :first) }
+        fill_in('link_url', with: 'http://angus.example.com/link-to-change')
+        click_on('Update')
+
+        expect(page).to have_text("Services and links")
+        expect(page).to have_link('http://angus.example.com/link-to-change')
```
whoojemaflip: Would this test pass if the page just reloaded?  What about an assertion on the URL?

app/controllers/links_controller.rb
```diff
@@ -1,20 +1,15 @@
 require 'local-links-manager/export/link_status_exporter'
 
 class LinksController < ApplicationController
-  before_action :load_dependencies
+  before_action :load_dependencies, :set_back_url_before_post_request
```
whoojemaflip: is this relevant to all controller methods, or could it be restricted to just one or two?

whoojemaflip (changes requested):

brenetic:
@whoojemaflip  Addressed all the comments and did most of the refactor for the `Links#index`. Unfortunately one set of failing tests remain . . . 💥 🔥 👎

issyl0:
Are the tests failing because there's a trailing `:` in `(def initialize(link, view_context:)` that it then can't find `view_context` without the `:` in the method itself? *shrug*

EDIT: Nope, that didn't help at all in dev. Confused (and I'd never seen that pattern before).

davidbasalla:
Could we split that out that last commit into a different PR? It seems like a substantial change which is not directly related to the 'edit link' page

brenetic:
@issyl0 the `view_context` is a named parameter

brenetic:
@davidbasalla  Removed the last commit. Will raise another PR

app/controllers/links_controller.rb
```diff
@@ -2,41 +2,48 @@
 
 class LinksController < ApplicationController
   before_action :load_dependencies
+  before_action :set_back_url_before_post_request, only: [:edit, :update, :destroy]
```
whoojemaflip, line 5: this might be more succinctly expressed as
`before_action :set_back_url, if: -> { request.post? }`

NBD though, only change it if you want to.

whoojemaflip (approved):
Cool cool cool

PR #94, opened 2016-11-28
Backup airbrake versions a bit

sihugh:
We're running an older version of errbit which doesn't seem to like the newer
versions of `airbrake` and `airbrake-ruby`.  These versions are the same as
Specialist-Publisher and Publishing-Api and worked OK when deployed in a
branch.


davidbasalla:
👍

PR #95, opened 2016-11-28
Update broken_link_count to ignore disabled links

davidbasalla:
Previously when calculating the broken links count, we would count links that were associated with disabled services. This led to a big discrepancy between link count and the listing of the broken links. This commit fixes the problem by using the `provided_service_links` method which takes enabled services and tiers into consideration. Note that we update the service factory to default to supporting all tiers, otherwise calling `provided_service_links` for a service (when created by Factory) would not return anything.

We also remove the `FactoryGirl.` call which is no longer necessary.


whoojemaflip:
🖖

PR #96, opened 2016-11-29
Remove the ability for users to edit homepage URLs

issyl0:
- This is being removed pending further research and design work,
  because from previous rounds of research, people didn't notice it, and
  homepages shouldn't change that often.


app/views/shared/_local_authority_details.html.erb
```diff
@@ -6,6 +6,5 @@
     <span class="text-muted"><%= authority.homepage_link_last_checked %></span>
   </p>
   <p>
```
whoojemaflip: does the parent `<p>` need to exist?

issyl0: Good spot. Removed.

whoojemaflip:
➿

PR #97, opened 2016-11-30
Fix caching issue

whoojemaflip:
The cache-keys for fragments rendered in the Specific Service and
Specific Council pages were incorrectly missing an identifier. This meant
that some services were occasionally appearing for unexpected councils.


sihugh:
:+1:

PR #102, opened 2016-12-05
Export bad link and homepage URLs and status

emmabeynon:
[Trello card](https://trello.com/c/SCWYMyWG/572-address-tim-s-llm-data-needs-3)

This adds functionality to export a CSV of non-200 status (aka "bad") link URLs and their status. A similar CSV is available for Local Authority homepages as well. The CSVs are available to download via URLs so that a performance analyst can access the data when required.

### DIFF
--- a/lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
+++ b/lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
@@ -1,0 +1,27 @@
+require 'csv'
+
+module LocalLinksManager
+  module Export
+    class BadLinksUrlAndStatusExporter
+      HEADINGS = %w(url status).freeze
+
+      def self.local_authority_bad_homepage_url_and_status_csv
+        CSV.generate do |csv|
+          csv << HEADINGS
+          LocalAuthority.where.not(status: "200").distinct.pluck(:homepage_url, :status).each do |la|
+            csv << la
+          end
+        end
+      end
+
+      def self.bad_links_url_and_status_csv
+        CSV.generate do |csv|
+          csv << HEADINGS
+          Link.enabled_links.currently_broken.distinct.pluck(:url, :status).each do |link|
+            csv << link
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
+++ b/spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+require 'local-links-manager/export/bad_links_url_and_status_exporter'
+
+describe LocalLinksManager::Export::BadLinksUrlAndStatusExporter do
+  let(:exporter) { LocalLinksManager::Export::BadLinksUrlAndStatusExporter }
+
+
+  describe ".local_authority_bad_homepage_url_and_status_csv" do
+    before do
+      create(:local_authority, homepage_url: "http://www.hogsmeade.gov.uk", status: "403")
+      create(:local_authority, homepage_url: "http://www.godricshollow.gov.uk", status: "Timeout Error")
+      create(:local_authority, homepage_url: "http://www.azkaban.gov.uk/", status: "Connection failed")
+      create(:local_authority, homepage_url: "http://www.littlehangleton.gov.uk", status: "404")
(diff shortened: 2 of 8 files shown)


spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+require 'local-links-manager/export/bad_links_url_and_status_exporter'
+
+describe LocalLinksManager::Export::BadLinksUrlAndStatusExporter do
+  let(:exporter) { LocalLinksManager::Export::BadLinksUrlAndStatusExporter }
+
+  describe ".local_authority_bad_homepage_url_and_status_csv" do
+    it "returns the URL and status for each local authority with a non-200 homepage" do
+      allow(LocalAuthority).to receive_message_chain("where.not.distinct.pluck").and_return([["http://www.hogsmeade.gov.uk", "403"], ["http://www.godricshollow.gov.uk", "Timeout Error"], ["http://www.azkaban.gov.uk/", "Connection failed"], ["http://www.littlehangleton.gov.uk", "404"]])
```
issyl0: 😍 I approve of your test data.

lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
```diff
@@ -0,0 +1,27 @@
...
+          LocalAuthority.where.not(status: "200").distinct.pluck(:homepage_url, :status).each do |la|
+            csv << la
+          end
+        end
+      end
+
+      def self.bad_links_url_and_status_csv
+        CSV.generate do |csv|
+          csv << HEADINGS
+          Link.enabled_links.where.not(status: "200").distinct.pluck(:url, :status).each do |link|
```
davidbasalla: We could use the `currently_broken` Link scope here

davidbasalla: `Link.enabled_links` actually returns a different number than if we looped through each LocalAuthority and called `provided_service_links` on it. The former currently returns '40895' in Integration, the latter '39105'. I think this is due to us not taking tiers into account for `enabled_links`. I've been meaning to have a look at this but not had a chance to. We should look at this at some point in the near future. It's probably fine for this PR, as the difference isn't very large.

spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
davidbasalla: Stubbing `"where.not.distinct.pluck"` feels a bit brittle. Could we try replacing this with creating some links via FactoryGirl?

lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
emmabeynon: Thanks - this is now using `currently_broken`.

spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
emmabeynon: Now using factories and tidied up the expectations to make them more readable 👍

spec/lib/local-links-manager/export/bad_links_url_and_status_exporter_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+require 'local-links-manager/export/bad_links_url_and_status_exporter'
+
+describe LocalLinksManager::Export::BadLinksUrlAndStatusExporter do
+  let(:exporter) { LocalLinksManager::Export::BadLinksUrlAndStatusExporter }
+
+  before do
+    create(:local_authority, homepage_url: "http://www.hogsmeade.gov.uk", status: "403")
```
davidbasalla: 💯 for these tests 😄 

A small thing, I'd probably move Link factory calls down to the tests for the links tests (into a separate `before` block). Should we also update 'links_status_exporter_spec.rb' to use FactoryGirl as well while we're at it?

emmabeynon: Thanks!  I've broken the before blocks out and updated links_status_exporter_spec.

davidbasalla:
Great! 👍

PR #103, opened 2016-12-05
Add rake task to remove ServiceTier duplicates

davidbasalla:
This allows us to remove duplicate entries for ServiceTiers. We accidently created these duplicated entries though our ServiceInteraction import rake task which runs at the beginning of each month. There will a separate PR for updating the import code.

We intend to remove this rake task once we fix the import code.


brenetic:
😰

PR #104, opened 2016-12-06
Fix service tier import duplication

emmabeynon:
Trello card: https://trello.com/c/IbaS4EjW/581-fix-servicetier-import-to-not-create-duplicates

This updates the ServiceTierImporter to not import duplicate ServiceTiers, and removes the rake task that deletes ServiceTier duplicates.


davidbasalla (changes requested):
I think there might still be an issue when the tiers for a Service change (eg from 'county/unitary' to 'district/unitary') which I don't think is handled by this code. The Service importer has a `create_or_update_record` method which (in concept) might be more appropriate here? 

Let's have a chat tomorrow as I'm a bit unclear on why we do this import from Publisher anyway. Feels like LocalLinksManager could be the source of truth for this data...?

emmabeynon:
Hi @davidbasalla, the first commit is now updated to handle services whose tiers change.

brenetic:
👍  This looks good to me. @davidbasalla has more context but FFTM

whoojemaflip:
It might be a good idea to add some validation to the `ServiceTier` model, which would ensure no duplicates could be created without throwing an error.

lib/local-links-manager/import/services_tier_importer.rb
```diff
@@ -41,33 +41,108 @@ def import_item(item, response, summariser)
...
         end
       end
 
-      def update_tier(service, tier_name)
+      def create_and_update_tiers(service, tier_name)
         case tier_name
         when 'district/unitary'
-          ServiceTier.create(service: service, tier_id: Tier.district)
-          ServiceTier.create(service: service, tier_id: Tier.unitary)
+          if service.tiers.include?('district')
```
whoojemaflip: This class is modifying data related to the `Service` class a lot. Maybe indicative of [feature envy](https://sourcemaking.com/refactoring/smells/feature-envy). Could this logic live in the `Service` class instead?

whoojemaflip (changes requested):

lib/local-links-manager/import/services_tier_importer.rb
```diff
@@ -41,34 +41,68 @@ def import_item(item, response, summariser)
...
 
         if item[:tier].blank?
           response.errors << "LGSL #{item[:lgsl_code]} is missing a tier"
           summariser.increment_ignored_items_count
-        elsif not update_tier(service, item[:tier])
+        elsif !service.valid_tier?(item[:tier])
           response.errors << "LGSL #{item[:lgsl_code]} has incorrect tier name"
           summariser.increment_ignored_items_count
         else
+          service.delete_and_create_tiers(item[:tier])
```
whoojemaflip, line 53: If this is run for every service, then all the `created_at` timestamps will be the same, right?

whoojemaflip:
is this missing a migration?

emmabeynon:
@whoojemaflip whoops, not sure where that went! Should be in the last commit now.

whoojemaflip (approved):

PR #105, opened 2016-12-07
Set test envars correctly for Jenkins2

brenetic:
Envars set in each Jenkins2 stage seem to not be available in other stages. This sets the `RAILS_ENV=test` in the test stage until this can be investigated further.

cc @emmabeynon


sihugh:
Seems fine :+1:

PR #106, opened 2017-01-25
Downgrade airbrake to 4.x and restore config

davidbasalla:
Our errbit installation doesn't like the airbrake 5.x gem and therefore deploys don't get notified properly. Downgrading to 4.x fixes this.

Inspired by https://github.com/alphagov/specialist-publisher/pull/988

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -40,9 +40,9 @@ GEM
       minitest (~> 5.1)
       tzinfo (~> 1.1)
     addressable (2.4.0)
-    airbrake (5.5.0)
-      airbrake-ruby (~> 1.5)
-    airbrake-ruby (1.5.0)
+    airbrake (4.3.8)
+      builder
+      multi_json
     airbrussh (1.1.1)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -369,5 +368,8 @@ DEPENDENCIES
   webmock (~> 1.2)
   whenever
 
+RUBY VERSION
```
whoojemaflip: I think this has snuck in there somehow

Gemfile.lock
```diff
@@ -369,5 +368,8 @@ DEPENDENCIES
...
+   ruby 2.2.3p173
+
 BUNDLED WITH
-   1.10.6
+   1.13.1
```
whoojemaflip: Can we avoid flip-flopping this value plz

whoojemaflip (changes requested):

davidbasalla:
@whoojemaflip thank you and fixed!

whoojemaflip (approved):

PR #109, opened 2017-04-05
Adds GA tracking

elenatanasoiu:
[Trello card](https://trello.com/c/GGvHKuad/118-event-tracking-and-analytics-for-iteration-2-of-local-links-manager-3)

Depends on changes being merged in https://github.com/alphagov/govuk_admin_template/pull/145 This will enable modules defined in govuk_admin_template to be used in local-links-manager.

We send out events (`track-submit`, `track-click`, `auto-track-event` etc) to GA.


app/views/local_authorities/index.html.erb
```diff
@@ -15,7 +15,10 @@
         <div class='filter-control'>
           <%= render partial: 'govuk_admin_template/table_filter_form', locals: { placeholder: 'Search for a council here' } %>
         </div>
-        <div class='count'>
+        <div class='count'
+             data-module="track-dimension"
+             data-track-index="Scorecard total",
```
fofr: Index should be a number.

elenatanasoiu: Changed to number

Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'airbrake', '~> 4.0'
 gem 'dalli'
 gem 'faraday'
 gem 'faraday_middleware'
-gem 'govuk_admin_template', '~> 4.2'
+gem 'govuk_admin_template'
```
boffbowsh: This should now be `~> 5.0`

Gemfile.lock
```diff
@@ -368,5 +377,8 @@ DEPENDENCIES
   webmock (~> 1.2)
   whenever
 
+RUBY VERSION
```
boffbowsh: Can you re-bundle this with your dev VM please? Otherwise people will get warnings on their own VMs until the ruby version this app uses is bumped.

boffbowsh (changes requested):
Couple of small changes. This looks mostly good though - can you get a review from @fofr please?

Gemfile
elenatanasoiu: Added

elenatanasoiu:
Asked @fofr for review

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+# Example: using a different boot file.
+# Jasmine.configure do |config|
+#    config.boot_dir = '/absolute/path/to/boot_dir'
+#    config.boot_files = lambda { ['/absolute/path/to/boot_dir/file.js'] }
+# end
+#
+# Example: prevent PhantomJS auto install, uses PhantomJS already on your path.
+Jasmine.configure do |config|
+  if ENV['TRAVIS']
+    config.prevent_phantom_js_auto_install = true
```
fofr: Why is this limited to `if ENV['TRAVIS']` and not applied all the time?

Other apps have this config set all the time:
eg https://github.com/alphagov/govuk_admin_template/blob/master/spec/javascripts/support/jasmine_helper.rb

elenatanasoiu: I've removed the environment check

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,3 @@
+//= require jquery
+//= require jquery_ujs
```
fofr: govuk_admin_template provides jQuery so the top line can be removed.
What is `jquery_ujs` used for?

app/assets/javascripts/modules/track_search.js
```diff
@@ -0,0 +1,24 @@
...
+  "use strict";
+
+  Modules.TrackSearch = function() {
+    var that = this;
+
+    that.start = function(container) {
+
+      var filterSearchInput = container.find('table'),
+          tableInput = container.find('.js-filter-table-input');
+      filterSearchInput.on('keyup', '.js-filter-table-input', trackSearch);
```
fofr: This will fire an analytics event after every keypress. What's the intention of this tracking? What are you trying to  find out?

app/views/interactions/index.html.erb
```diff
@@ -43,7 +43,13 @@
           </p>
         </td>
         <td class='text-center'>
-          <%= link_to(interaction.button_text, edit_link_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s") %>
+          <div class='actions add-vertical-margins col-xs-6 no-gutter'
```
fofr: Minor: inconsistent attribute quotes

app/views/interactions/index.html.erb
```diff
@@ -43,7 +43,13 @@
...
+               data-module="track-click"
+               data-track-action="editButtonClick"
+               data-track-label="<%= edit_link_path(interaction_slug: interaction.slug) %>"
+               data-track-category="userInteraction:LLM">
+               <%= link_to(interaction.button_text, edit_link_path(interaction_slug: interaction.slug), class: "btn btn-default btn-s js-track") %>
```
fofr: I'm not sure this will track as you intended. The module pulls the label and value from the link, and the action and category from the container. This will use the text of the link as the label, rather than the one provided in the attribute.

See: https://github.com/alphagov/govuk_admin_template/pull/145/files#diff-3018827fa99e362685623cba2c52c238

What are the styles on the wrapping `div` necessary for? If the layout has changed, could you include screenshots?

app/views/links/edit.html.erb
```diff
@@ -1,38 +1,53 @@
 <% breadcrumb :links, @local_authority, @service, @interaction %>
 
-<div class="page-title">
-  <h1><%= @local_authority.name %></h1>
+  <div class="page-title">
```
fofr: The new indentation here is wrong.

app/views/local_authorities/index.html.erb
```diff
@@ -8,30 +8,41 @@
...
+    <thead>
+      <tr class="table-header">
+        <th>
+          <div class='filter-control'>
+            <%= render partial: 'govuk_admin_template/table_filter_form', locals: { placeholder: 'Search for a council here' } %>
           </div>
-        </td>
+
+          <div class="count">
+            <% content_for :custom_dimension_js do %>
```
fofr: Why is this template variable being set here? I'd expect it to be set somewhere further up, at the start of a template or elsewhere.

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+#
+# Example: using a different boot file.
+# Jasmine.configure do |config|
+#    config.boot_dir = '/absolute/path/to/boot_dir'
+#    config.boot_files = lambda { ['/absolute/path/to/boot_dir/file.js'] }
+# end
+#
+# Example: prevent PhantomJS auto install, uses PhantomJS already on your path.
+Jasmine.configure do |config|
+  config.prevent_phantom_js_auto_install = true
```
fofr, line 13: :+1:

app/assets/javascripts/modules/track_search.js
```diff
@@ -0,0 +1,24 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.TrackSearch = function() {
+    var that = this;
+
+    that.start = function(container) {
+
+      var filterSearchInput = container.find('table'),
```
fofr: A table is not a`filterSearchInput`.

The two var declarations and the event listener can be reduced to:
`container.on('keyup', '.js-filter-table-input', trackSearch);`

app/assets/javascripts/modules/track_search.js
```diff
@@ -0,0 +1,24 @@
...
+          tableInput = container.find('.js-filter-table-input');
+      filterSearchInput.on('keyup', '.js-filter-table-input', trackSearch);
+
+      function trackSearch() {
+        var action = container.data('track-action'),
+          label = container.data('track-label'),
+          category = container.data('track-category'),
+          value = $.trim(tableInput.val());
```
fofr: Values must be numerical, this won't work.

fofr (changes requested):
I've found some significant problems with this PR. Specifically, it's not clear what the filter input analytics intends to track – it's sending a lot of events and the data attributes passed to it contain mistakes.

Was this tested before the PR was raised?

app/assets/javascripts/modules/track_search.js
elenatanasoiu: I've renamed tableInput to filterSearchInput and removed the original var: 
`filterSearchInput = container.find('.js-filter-table-input');`

I've replaced `filterSearchInput.on('keyup', '.js-filter-table-input', trackSearch);` with `container.on('keyup', '.js-filter-table-input', trackSearch);`

However, I still need the new `filterSearchInput` in order to get the text being searched for, so I've kept that variable.

app/assets/javascripts/modules/track_search.js
elenatanasoiu: I've moved the text being search to the label variable.

app/views/interactions/index.html.erb
elenatanasoiu: Fixed wherever I could find it.

app/views/interactions/index.html.erb
elenatanasoiu: I've removed the div wrapping. And I've moved the label as a property on the link.

elenatanasoiu: This is what the div wrapping was for (red Delete button)
screen shot 2017-04-12 at 17 49 36

This wasn't necessary on the index pages which looks like this (Edit button)
screen shot 2017-04-12 at 17 56 48

app/views/links/edit.html.erb
elenatanasoiu: Fixed

app/views/local_authorities/index.html.erb
elenatanasoiu: I've moved this up to the top of the page

app/assets/javascripts/application.js
elenatanasoiu: I've removed them. I believe I was using them to check jasmine tests in the browser, but they are redundant, I agree.

app/assets/javascripts/modules/track_search.js
elenatanasoiu: Ideally it's to send what's being searched back to GA. However, we could just fire that the filter was being used. 

I'll wait to talk to Tim LB before continuing with this, as it might be dropped.

fofr:
This advanced analytics tracking is no longer required.

PR #110, opened 2017-04-10
Add Google Analytics download code

sihugh:
I've rebased this on top of https://github.com/alphagov/local-links-manager/pull/113 so we should review that first

We want to get analytics for external links that are clicked on within local
transactions.  The intention is that we will use these to rank links in local
links manager.  More clicks means that the link is more important and should
be prioritised.

Running the GA code locally can be done by creating a file at
`config/local_env.yml`.  This should contain values for `GOOGLE_PRIVATE_KEY`,
`GOOGLE_CLIENT_EMAIL` and `GOOGLE_ANALYTICS_GOVUK_VIEW_ID`.  You can either
grab an example file from me, or construct your own with other credentials.
These values are set as environment variables in our non-Dev environments.

The import task adds the analytics count to each link. Not all links will
appear in GA.  In addition, it's possible that our query will return things
that we're not interested in.
When we're confident that we've had a successful import, we want to reset the
count on the other links to zero to ensure that we're only counting the last
week.

We don't want to use the local link to do the lookup, as it's possible that
this has changed over the week.  We're interested in the popularity of a
ServiceInteraction for a Local Authority, not any particular URL.

The GA client is a similar approach to that used by content-performance-manager


db/migrate/20170410075304_add_govuk_slug_and_title_to_service_interactions.rb
```diff
@@ -0,0 +1,9 @@
+class AddGovukSlugAndTitleToServiceInteractions < ActiveRecord::Migration[5.0]
```
issyl0: I don't think you meant to put this file in here, it was in #113.

Gemfile.lock
```diff
@@ -368,5 +407,8 @@ DEPENDENCIES
   webmock (~> 1.2)
   whenever
 
+RUBY VERSION
```
issyl0: Do the following new/changed lines have consequences elsewhere?

lib/local-links-manager/import/publishing_api_importer.rb
```diff
@@ -0,0 +1,94 @@
+require_relative 'errors'
```
issyl0: I think your rebase got a bit confused.

issyl0 (changes requested):

lib/tasks/import/google_analytics.rake
```diff
@@ -0,0 +1,11 @@
+require 'local-links-manager/import/analytics_importer'
+
+namespace :import do
+  desc "Imports analytics so that links can be prioritised by usage"
+  task google_analytics: :environment do
+    service_desc = 'Import Google Analytics to Local Links Manager'
+    response = LocalLinksManager::Import::AnalyticsImporter.import
+    Services.icinga_check(service_desc, response.successful?, response.message)
+    puts response.message
```
issyl0, line 9: Do we still needs this `puts`?

issyl0 (approved):
Lots of code, but it seems to be doing the right thing. I didn't realise GA could be so complicated.

PR #111, opened 2017-04-10
Integrate Link Checker API

emmabeynon:
Paired with @thomasleese 

[Trello Card](https://trello.com/c/QljohdgW/968-complete-local-links-manager-integration)


spec/factories/links.rb
```diff
@@ -4,6 +4,8 @@
     association :service_interaction
     sequence(:url) { |n| "http://www.example.com/#{n}" }
     status nil
+    link_errors ""
+    link_warnings ""
```
kevindew: Should these be empty hashes?

lib/local-links-manager/check_links/link_status_requester.rb
```diff
@@ -0,0 +1,31 @@
+module LocalLinksManager
+  module CheckLinks
+    class LinkStatusRequester
+      #FIX-ME make this load from routes
+      CALLBACK_URI = Plek.find('local-links-manager') + '/link-check-callback'
```
kevindew: it'd probably make most sense to inject this url in via the initializer. if you set an `as: :link_checker_callback` or similar you can use `app.link_checker_callback_url` - I guess if this is run from a rake context you'll probably have to configure a default hostname.

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+require 'local-links-manager/check_links/link_status_requester'
```
kevindew: is there an autoloading issue?

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+describe LocalLinksManager::CheckLinks::LinkStatusRequester do
+  subject(:link_status_requester) { described_class.new }
+
+  context "links for enabled services" do
+    let(:local_authority_1) { FactoryGirl.create(:local_authority) }
+    let(:local_authority_2) { FactoryGirl.create(:local_authority) }
+    let!(:link_1) { FactoryGirl.create(:link, local_authority: local_authority_1, url: 'http://www.example.com') }
+    let!(:link_2) { FactoryGirl.create(:link, local_authority: local_authority_2, url: 'http://www.example.com/exampl.html') }
```
kevindew: typo or intentional?

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    it 'does not test links' do
+      stub = stub_request(:post, "http://link-checker-api.dev.gov.uk/batch")
+      .with(body: request_body(disabled_service_link.local_authority.homepage_url).to_json)
+      .to_return(status: 200)
+
+      link_status_requester.call
+      expect(stub).to have_been_requested
+    end
+  end
```
kevindew: I'm a bit confused here - which is probably my brain being the problem. Are we testing here that it only tests the homepage url for a disabled service link and not other links?

If so, should we be setting up a local auth and some links for it to not have?

spec/requests/link_checker_api_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+require 'local-links-manager/check_links/link_status_updater'
+
+describe LocalLinksManager::CheckLinks::LinkStatusUpdater, type: :request do
+  subject(:status_updater) { described_class.new }
+
+  before do
+    @time = Timecop.freeze('2016-06-21 09:26:56 +0100')
+  end
```
kevindew: I think this would make more sense within the context

spec/requests/link_checker_api_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+                http_client_error: "Received 4xx response"
+              },
+              warnings: {
+                http_non_200: "Page not available."
+              },
+            }
+          ]
+        }
+
+        post '/link-check-callback', params: payload.to_json
```
kevindew: This probably also makes sense in the before block which would make this `it` block a lot shorter

spec/requests/link_checker_api_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
```
kevindew: We've got a bit of mix of single and double quotes on the strings in this file

app/models/link.rb
```diff
@@ -18,10 +18,8 @@ class Link < ApplicationRecord
       .where(service_interactions: { service_id: service })
   }
 
-  HTTP_OK_STATUS_CODE = 200
-
-  scope :good_links, -> { where(status: HTTP_OK_STATUS_CODE) }
-  scope :currently_broken, -> { where.not(status: HTTP_OK_STATUS_CODE) }
+  scope :good_links, -> { where(status: "ok") }
+  scope :currently_broken, -> { where.not(status: "ok") }
```
kevindew: I'd think we only consider links with a status of broken as `currently_broken` Would have thought caution would be a separate grouping. Or if it'll be just one grouping it could be "For Review" rather than "Broken"? What do you think? Maybe Mia will have thoughts?

app/models/link.rb
```diff
@@ -60,7 +58,12 @@ def link_with_matching_url
   end
 
   def existing_link_url
-    @_link ||= Link.where(url: self.url).distinct.first
+    link = Link.where(url: self.url).group_by(&:url)[self.url]
+    if link
+      @_link ||= link.first
+    else
+      @_link
+    end
```
kevindew: I'm a bit confused about this (both versions of it) can we chat about it?

lib/local-links-manager/check_links/link_checker.rb
```diff
@@ -1,46 +0,0 @@
-module LocalLinksManager
```
kevindew, line 1: Yay all the code going :tada:

kevindew (commented):
Looking really good 👍  - a bunch of comments included, the outdated ones still seem to apply so left them in there

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
thomasleese: Local links manager doesn't seem to have autoloading. I would say out of the scope of this PR?

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
thomasleese: Typo

spec/requests/link_checker_api_spec.rb
thomasleese: Local links manager doesn't seem to be as strict on the quotes as `publishing-api`...

kevindew:
@thomasleese you might want to cherry pick out the autoloading commit and put that in as a separate PR so that can get merged early and the benchmarking team can use it - seems like it's a likely source of merge conflicts.

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
emmabeynon: Now fixed ⚡️

app/models/link.rb
emmabeynon: This had to be changed as a result of `link_errors` and `link_warnings` storing JSON.  After this change was made, the `distinct` method stopped working due to a Postgres error `could not identify an equality operator for type json`.  After doing some research I found this could be fixed by changing the attributes' type to jsonb, however this requires a newer version of Postgres, so I used `group_by` as a workaround to avoid having to upgrade Postgres. Not very pretty, but it does the job!

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
emmabeynon: The link factory has a local authority association, so a local authority (and also homepage) is automatically created for the link :) However, we probably should test that the disabled link is not included in the batch, so I'll add that in.

lib/tasks/check-links/link_checker.rake
```diff
@@ -1,31 +1,20 @@
...
 
 desc "Check links"
 task "check-links": :environment do
   service_desc = "Local Links Manager link checker rake task"
-  LocalLinksManager::DistributedLock.new('check-links').lock(
-    lock_obtained: ->() {
-      begin
-        Rails.logger.info("Lock obtained, starting link checker")
-        Services.icinga_check(service_desc, true, "Lock obtained, starting link checker")
+  begin
```
sihugh: What was the reasoning behind removing the distributed lock?  We still only want to run this on one machine don't we?

thomasleese: Hmm, perhaps I misunderstood what the point of the lock was, I figured it was there to avoid someone accidentally running the take command twice. Since the rake task now only takes a few minutes (it sends off requests to the `link-checker-api` which then responds through a web hook) is it necessary to have the lock?

sihugh: It's less the amount of time the request takes, rather that we probably don't want to update all the links multiple times if we can avoid it.

thomasleese: That shouldn't be a problem. The Link Checker API automatically caches results for 24 hours for us anyway, and we only need to perform the update queries if the last checked time has been updated.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  before_action :verify_signature
+
+  def link_check_callback
+    LocalLinksManager::CheckLinks::LinkStatusUpdater.new.call(
+      GdsApi::LinkCheckerApi::BatchReport.new(
+        params.to_unsafe_hash
+      )
+    )
+  end
+
```
kevindew, line 16: Should probably have these as private methods

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  def verify_signature
+    given_signature = request.env["HTTP_X_LINKCHECKERAPI_SIGNATURE"]
```
kevindew: can we use a request.headers method? looking at [source](https://github.com/rails/rails/blob/d2dde6cbfc11bf39be30d22e2ee7c41c119e0d51/actionpack/lib/action_dispatch/http/request.rb#L188) they header might be formatted in header spec there (or maybe it's a nightmare because they format them different)

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    return head :bad_request unless given_signature
+
+    request.body.rewind
+    body = request.body.read
```
kevindew: rather than needing to do this can we just use `request.raw_post` ?

app/models/link.rb
```diff
@@ -56,31 +58,35 @@ def self.build(params)
...
 
   def link_with_matching_url
-    existing_link_url || existing_homepage_url
+    existing_link || existing_homepage
   end
 
-  def existing_link_url
-    @_link ||= Link.where(url: self.url).distinct.first
+  def existing_link
+    @_link ||= Link.where(url: self.url).group_by(&:url)[self.url].try(:first)
```
kevindew: Do we gain anything here over doing `Link.where(url: self.url).first`?

app/presenters/url_status_presentation.rb
```diff
@@ -3,17 +3,26 @@ module UrlStatusPresentation
...
-    return "Good" if status == '200'
-    return "Broken Link #{status}" if status.start_with?('4')
-    return "Server Error #{status}" if status.start_with?('5')
-    status
+    return "Good" if status == "ok"
+    if status == "broken"
+      "Broken: #{link_errors.keys.first}"
+    elsif status == "caution"
+      "Note: #{link_warnings.keys.first}"
+    else
```
kevindew: Might we be dealing with pending links?

lib/local-links-manager/check_links/link_status_requester.rb
```diff
@@ -0,0 +1,42 @@
+require "gds_api/link_checker_api"
+
+module LocalLinksManager
+  module CheckLinks
+    class LinkStatusRequester
+      delegate :url_helpers, to: "Rails.application.routes"
```
kevindew, line 6: 🎉 🌮

lib/local-links-manager/check_links/link_status_updater.rb
```diff
@@ -1,40 +1,43 @@
...
+
+        Link
+          .where(url: link_report.uri)
+          .last_checked_before(link_report.checked)
+          .update_all(**fields)
+
+        LocalAuthority
+          .where(homepage_url: link_report.uri)
+          .link_last_checked_before(link_report.checked)
+          .update_all(**fields)
```
kevindew: Do these need the double splat operator? would have thought they had no effect

kevindew (commented):
all looking good 👍

app/models/link.rb
emmabeynon: No, I think you're right.  Will change!

app/presenters/url_status_presentation.rb
emmabeynon: I don't think we'll be sending back any pending links...am I correct in thinking that, @thomasleese?

kevindew: Sure, I guess if your only source is webhook and that's not firing until complete you won't get any pendings

lib/local-links-manager/check_links/link_status_updater.rb
emmabeynon: Tests pass without the double splat operator, so will remove them ⚡️

app/presenters/url_status_presentation.rb
thomasleese: In my head, I had thought we were ignoring links when they came back as pending, but I'm no longer sure that's the case, would be worth checking.

thomasleese: Yes, what @kevindew said makes more sense than what I did!

lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
```diff
@@ -3,22 +3,22 @@
...
-      HEADINGS = %w(url status).freeze
-
+      HEADINGS = %w(url link_errors link_warnings).freeze
       def self.local_authority_bad_homepage_url_and_status_csv
         CSV.generate do |csv|
           csv << HEADINGS
-          LocalAuthority.where.not(status: "200").distinct.pluck(:homepage_url, :status).each do |la|
-            csv << la
+          LocalAuthority.where.not(status: "ok").pluck(:homepage_url, :link_errors, :link_warnings).each do |la|
+            csv << [la[0], la[1].join(','), la[2].join(',')]
```
kevindew: not sure about the wisdom of using comma separators here since we have commas in some of the messages. I wonder if we could just have status and problem summary and not worry about the lists?

kevindew (approved):
I've not found anything else to nag about so I'm going to say LGTM 👍  😀  - probably worth getting a look from someone on the team that owns this

brenetic:
🚀  Looks good!! 🚀

PR #113, opened 2017-04-12
Import titles and slugs of local transactions

sihugh:
We're going to present the title of local transactions above the links which
will be shown on that transaction.  This will help administrators to choose a
more accurate link to match the content on the GOV.UK page.

The slug is to be used to provide a link to the local transaction page so that
an adminstrator can navigate directly to the page and check the body of the
content if necessary.

The slug and the title are both assigned to a local transaction page which has
a combination of LGIL and LGSL code.  This maps closely to the concept of a
service interaction which is why we've added them to that model.

The slug is also going to be used to marry up Google analytics for each page/

We're also tracking which service interactions are actually live on GOV.UK.  We
have a lot of links which correspond to LGSL/LGIL combinations that aren't
ever used.  By using this import we can check which ones are used for GOV.UK,
and only show those ones to administrators.  We can also delete the unused ones
as they will not be curated.

We should track if we consider a service to be enabled, and it then disappears
from the import.  This could be down to user error, or the retiring of a local
transaction.  Either way, we want to know about it - which is why an error will
be raised.

The fallback code will be removed once all local transactions have an LGIL code
assigned.

### DIFF
--- a/lib/local-links-manager/import/publishing_api_importer.rb
+++ b/lib/local-links-manager/import/publishing_api_importer.rb
@@ -1,0 +1,94 @@
+require_relative 'errors'
+require_relative 'error_message_formatter'
+require_relative 'import_comparer'
+require_relative 'processor'
+require 'gds_api/publishing_api_v2'
+
+module LocalLinksManager
+  module Import
+    class PublishingApiImporter
+      def self.import
+        new.import_data
+      end
+
+      def initialize(import_comparer = ImportComparer.new)
+        @comparer = import_comparer
+      end
+
+      def import_data
+        Processor.new(self).process
+      end
+
+      def each_item(&block)
+        local_transactions.each(&block)
+      end
+
+      def import_item(local_transaction, _response, summariser)
+        raise MissingIdentifierError, "Found empty LGSL/LGIL code on local_transaction #{local_transaction['slug']}" if local_transaction["lgil"].blank? || local_transaction["lgsl"].blank?
+
+        service = Service.find_by(lgsl_code: local_transaction["lgsl"])
+        interaction = Interaction.find_by(lgil_code: local_transaction["lgil"])
+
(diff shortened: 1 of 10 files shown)


spec/lib/local-links-manager/import/publishing_api_importer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+describe LocalLinksManager::Import::PublishingApiImporter do
+  include GdsApi::TestHelpers::PublishingApiV2
+
+  describe 'import of slugs and titles from Publishing API' do
+    context 'when Publishing API returns Local Transactions' do
+      let(:local_transaction) {
+        {
+          "base_path" => "/ring-disposal-services",
+          "description" => "Contact the council of Elrond to discuss disposing of powerful magic rings",
```
issyl0, line 13: 👏

lib/local-links-manager/import/publishing_api_importer.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      def publishing_api_response
+        Services.publishing_api.get_content_items(document_type: 'local_transaction', per_page: 150)
+      end
+
+      def local_transaction_hash(parsed_result)
+        local_transaction = {}
+        local_transaction["title"] = parsed_result["title"]
+        local_transaction["slug"] = parsed_result["base_path"][1..-1]
```
issyl0, line 83: Why? Is this to strip the leading / from the slug?

issyl0 (approved):
Seems sensible to me, especially all the error handling for if they disappear.

lib/local-links-manager/import/publishing_api_importer.rb
sihugh: Yes! :-)

PR #118, opened 2017-05-23
Schedule import of local transactions GA data

sihugh:
Scheduling the analytics import so that we have up to date usage data on our links.

We don't want to reset the analytics count on every link that is not in the GA
import because most of them are already at zero.  Instead, we'll make a note
of which links have a non-zero count beforehand, and just reset those that
haven't been updated in the import.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -16,3 +16,9 @@ end
 every :day, at: '3am' do
   rake 'export:links:all'
 end
+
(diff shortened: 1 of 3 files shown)


issyl0:
🔴 INCONSISTENT INDENTATION DETECTED, reports the linter.

issyl0 (approved):

sihugh:
Spectrum Is Green :-)

PR #121, opened 2017-07-11
Change link checking strategy

sihugh:
If we submit all the links for a local authority at once, we have noticed that
several authorities fail link checking for lots of their links.  I guess that
this is because they haven't warmed up sufficiently at 2AM when the link
checker runs.  We run 4 threads on the sidekiq worker (in link checker api) 
at present and it's possible that we are causing our own bottleneck on the 
websites we are checking.  (Some of the links are very slow to respond)

I propose a different strategy where we check links by service instead.  This
spaces out the requests to local authority websites a bit which allows them
extra time to warm up.

I've also changed the order so that the least significant (by GA click count)
is requested first.  I'm hoping that this will contribute to a more accurate
link checker status for the high value links in local links manager which is
what we're after.

We check the home pages last of all as it's most important that they are correct.

I've also added a rake task allowing us to recheck an individual council, as
sometimes they are down for maintenance overnight. There is nothing that our
admins can do, so we should be able to easily recheck.

### DIFF
--- a/lib/local-links-manager/check_links/link_status_requester.rb
+++ b/lib/local-links-manager/check_links/link_status_requester.rb
@@ -6,16 +6,35 @@ module LocalLinksManager
       delegate :url_helpers, to: "Rails.application.routes"
 
       def call
-        LocalAuthority.all.each do |local_authority|
-          link_checker_api.create_batch(
-            urls_for_local_authority(local_authority),
-            webhook_uri: webhook_uri,
-            webhook_secret_token: webhook_secret_token
-          )
+        Service.enabled.each do |service|
+          check_urls service.links.order(analytics: :asc).map(&:url).uniq
         end
+
+        check_urls homepage_urls
+      end
+
+      def check_authority_urls(authority_slug)
+        check_urls urls_for_authority(authority_slug).uniq
       end
 
     private
+
+      def urls_for_authority(authority_slug)
+        local_authority = LocalAuthority.find_by(slug: authority_slug)
+        local_authority.links.map(&:url) << local_authority.homepage_url
+      end
+
+      def homepage_urls
+        LocalAuthority.all.map(&:homepage_url)
+      end
+
+      def check_urls(urls)
+        link_checker_api.create_batch(
+          urls,
+          webhook_uri: webhook_uri,
+          webhook_secret_token: webhook_secret_token
+        )
+      end
 
       def webhook_uri
         Plek.find("local-links-manager") + url_helpers.link_checker_webhook_path
@@ -23,11 +42,6 @@ module LocalLinksManager
 
       def webhook_secret_token
         Rails.application.secrets.link_checker_api_secret_token
-      end
-
-      def urls_for_local_authority(local_authority)
-        (local_authority.provided_service_links.map(&:url) +
-          [local_authority.homepage_url]).uniq
       end
 
       def link_checker_api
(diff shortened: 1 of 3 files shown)


lib/local-links-manager/check_links/link_status_requester.rb
```diff
@@ -6,17 +6,36 @@ class LinkStatusRequester
       delegate :url_helpers, to: "Rails.application.routes"
 
       def call
-        LocalAuthority.all.each do |local_authority|
-          link_checker_api.create_batch(
-            urls_for_local_authority(local_authority),
-            webhook_uri: webhook_uri,
-            webhook_secret_token: webhook_secret_token
-          )
+        Service.where(enabled: true).each do |service|
```
emmabeynon: I think we can just do `Service.enabled.each` here. Hurrah for scopes :)

lib/local-links-manager/check_links/link_status_requester.rb
```diff
@@ -6,17 +6,36 @@ class LinkStatusRequester
       delegate :url_helpers, to: "Rails.application.routes"
 
       def call
-        LocalAuthority.all.each do |local_authority|
-          link_checker_api.create_batch(
-            urls_for_local_authority(local_authority),
```
emmabeynon, line 11: We should probably remove this method from line 47, as it's no longer being used.

lib/tasks/check-links/link_checker.rake
```diff
@@ -25,3 +25,11 @@ task "check-links": :environment do
     }
   )
 end
+
+namespace :"check-links" do
+  desc "Check links for a single local authority"
+  task :local_authority, [:authority_slug] => :environment do
+    checker = LocalLinksManager::CheckLinks::LinkStatusRequester.new
```
emmabeynon, line 32: Do we need to add a lock here, like how we have for checking all links?

sihugh: If we add it to a schedule then I would support that, but it'll be used for ad-hoc re-checks at the moment.

lib/local-links-manager/check_links/link_status_requester.rb
sihugh: That's a good spot, I hadn't realised that I'd rewritten it.

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
```diff
@@ -8,20 +8,26 @@
   subject(:link_status_requester) { described_class.new }
 
   context "links for enabled services" do
```
emmabeynon: Perhaps we should make it explicit in the context that we're checking homepage URLs as well?

emmabeynon:
All looks good to me, just a couple of comments. Perhaps @thomasleese would like a quick review in case there's some consideration we're not aware of with the change in strategy?

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
sihugh: I can see why this doesn't say `homepage_urls`, as the difference between this context and the next is that the services are enabled or disabled.  Maybe we should name one `with homepage urls and links for enabled services` and the other `with homepage urls and links for disabled services`?

thomasleese:
Yep, this new strategy seems like a good idea to me!

spec/lib/local-links-manager/check_links/link_status_requester_spec.rb
emmabeynon: Sounds reasonable to me 👍

lib/local-links-manager/check_links/link_status_requester.rb
sihugh: Removed now

emmabeynon (approved):

sihugh:
Thanks both!

PR #125, opened 2017-08-25
Don't lint sass on build temporarily

sihugh:
We're currently not able to build master because of this.  We've got a card for
fixing the errors here: https://trello.com/c/9ppiSBt8/140-fix-local-links-manager-build-on-master


sihugh:
Closed in favour of actually fixing ( https://github.com/alphagov/local-links-manager/pull/126 )

PR #129, opened 2017-09-26
Allow Links with null urls

sihugh:
In this PR we enable having Link items with null urls so that we can prioritise the fixing of missing links by using analytics of users requesting them through local transactions on www.gov.uk

We want to ensure that they won't be exposed in:
  - API reponses
  - Link checker requests
  - The admin UI
  - The links export
This allows us to add the capability to manage missing links to the admin UI bit by bit without worrying that we're impacting other elements of the system.

We then add a rake task to add the missing links.  This would need to be re-run should we add further local transactions on GOV.UK


lib/local-links-manager/check_links/link_status_requester.rb
```diff
@@ -24,11 +24,11 @@ def check_authority_urls(authority_slug)
 
       def urls_for_authority(authority_slug)
         local_authority = LocalAuthority.find_by(slug: authority_slug)
-        local_authority.links.map(&:url) << local_authority.homepage_url
+        local_authority.links.with_url.map(&:url) << local_authority.homepage_url
       end
 
       def homepage_urls
-        LocalAuthority.all.map(&:homepage_url)
+        LocalAuthority.all.map(&:homepage_url).sort
```
brenetic, line 31: Does the link checker require them sorted?

brenetic:
🚀  LGTM! :shipit:

lib/local-links-manager/check_links/link_status_requester.rb
sihugh: It doesn't but I wanted to have them in a deterministic order, partly for my OCD and partly so the test didn't randomly fail.  I could have been a little more overt about this 🍇

PR #130, opened 2017-09-28
Allow editing of missing links

sihugh:
In this PR we enable showing missing links on the three relevant pages: the homepage, services pages and local authority pages.

We tweak the rake task to add missing links so we ensure that a status is set and prevent links from being deleted in the UI (as we'd only have to run the rake task again)

We also remove the "Good Links" tab as it is never used and the "Broken" and "All" tabs are sufficient.

I've also updated the breadcrumbs as the hierarchy was a bit short before.  Now we have a way to navigate from a service to a local authority and back

Screenshots of missing links on each page are below.  We'll update the local authority and service pages in subsequent PRs

Services page
screen shot 2017-09-28 at 13 57 07

Local authority page
screen shot 2017-09-28 at 13 56 46

Home page
screen shot 2017-09-28 at 13 56 30

### DIFF
--- a/app/controllers/local_authorities_controller.rb
+++ b/app/controllers/local_authorities_controller.rb
@@ -1,6 +1,8 @@
 require 'local-links-manager/export/bad_links_url_and_status_exporter'
 
 class LocalAuthoritiesController < ApplicationController
+  include LinkFilterHelper
+
   def index
     @authorities = LocalAuthority.order(broken_link_count: :desc)
     raise RuntimeError.new('Missing Data') if @authorities.empty?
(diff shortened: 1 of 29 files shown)


app/controllers/local_authorities_controller.rb
```diff
@@ -32,9 +32,7 @@ def filtered_links
 
     case params[:filter]
     when 'broken_links'
-      links.currently_broken
-    when 'good_links'
-      links.good_links
+      links.broken_or_missing
```
Davidslv: Seems there's a repetition of code here (with services_controller), we could extract this into another helper class to receive the `params[:filter]` and return the "right thing"

PR #131, opened 2017-10-18
Fix import command to match rake task

sihugh:
It seems that we've been using the wrong name for the rake task so this
hasn't been running :-(

https://github.com/alphagov/local-links-manager/blob/776000408cde81b276105479e760c7fab9db9bbc/lib/tasks/import/google_analytics.rake#L5

Part of: https://trello.com/c/hwYmLdBN/229-fix-local-links-manager-ga-import-and-run-more-often


deborahchua (approved):
👍

PR #132, opened 2017-11-02
Bad links exporter

elenatanasoiu:
For [Trello card](https://trello.com/c/SYgBUl4q/258-develop-an-upload-to-ga-of-link-checker-data)

## WHAT

This PR makes a list of non-200 status links (aka "bad links") available to GA by generating a CSV file with these links and then uploading it to Google Analytics as a Custom Import. We're using the Google Analytics Management API to do this. 

You will need a `local_env.yaml` file for testing the upload locally. Be advised that the data being uploaded to Google Analytics will affect the data that already exists there. Please only upload empty data when testing, like so: 

```
data = "ga:dimension36,ga:dimension37
'',''
"
```

The app assumes five environment variables are made available through `govuk-secrets`: 
- `GOOGLE_CLIENT_EMAIL` (this is extracted from the JSON file containing the credentials for your service account, in step 1)
- `GOOGLE_PRIVATE_KEY` (this is extracted from the JSON file containing the credentials for your service account, in step 1)
- `GOOGLE_EXPORT_ACCOUNT_ID` (this is obtained from the GA account domain you are trying to upload to. We use the account id for `www.gov.uk`)
- `GOOGLE_EXPORT_TRACKER_ID` (this is the web property UA string associated with the upload. Ex: `UA-XXXXXXXXX-X`)
- `GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID` (this is obtained from the custom data import setup in GA. Ex: `abcdeFGHijk12_AB12cd`)

More info in `docs/google_analytics_custom_import.md`. An update to the docs was added in a later PR: https://github.com/alphagov/local-links-manager/pull/134

That update described how you can only use one service account at a time, so there's no point in creating multiple ones. A workaround is recommended in the updated version of the docs. 

## WHY
 
The final goal is to make data from local-links-manager to non-specialists, aka in our case people who are using GA. The data can then be scheduled to be uploaded once a week. 

## CAVEATS

During the development of this PR, we wanted to take the CSV data and join it with existing hit data that we have on links. The links would be stored in the "event action" field. 
We discovered two things:
1. You can't do joins on that particular field, only on data in custom dimensions. The GA interface is designed in such a way that you can only pick custom dimensions when creating joins. 
2. It's not possible to create a filter in GA to duplicate the "event_action" data into a custom dimension. The filter will be applied before the data from the upload is loaded, so the filter won't know which data to join with as it won't be present when it uploads. 

The workaround for this will be, moving forward, to report the url in a custom dimension when we're sending hit data. 

## Further reading 

- Google Apps Service accounts: https://developers.google.com/api-client-library/ruby/auth/service-accounts
- Authorization levels for the GA Management API: https://developers.google.com/analytics/devguides/config/mgmt/v3/authorization
- Setting up a custom data import: https://developers.google.com/analytics/devguides/config/mgmt/v3/data-import
- Content data import example: https://support.google.com/analytics/answer/3271771
- Google API Client Library (Ruby): https://developers.google.com/api-client-library/ruby/start/get_started

## GA Config

The custom data import is set up as under the name "External Link status" in GA.


lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
```diff
@@ -22,6 +23,15 @@ def self.bad_links_url_and_status_csv
           end
         end
       end
+
+      def self.ga_bad_links_url_and_status_csv
```
h-lame: Given we've tied the these methods together in the specs, could we tie them together in implementation too?  Perhaps a private bad_links_csv method that takes a set of headings?  I don't know how strongly I feel about doing this though - I'm not sure if *they* are the same report with different headings, or just co-incidentally are the same report right now.

lib/google_analytics/analytics_export_service.rb
```diff
@@ -0,0 +1,31 @@
...
+module GoogleAnalytics
+  class AnalyticsExportService
+    include Google::Apis::AnalyticsV3
+    include Google::Auth
+
+    attr_accessor :service, :account_id, :web_property_id, :custom_data_source_id
+
+    def build
+      @service = AnalyticsService.new
+      @service.authorization ||= ServiceAccountCredentials.make_creds(
```
h-lame: We might be able to avoid storing the `.json` file.  We were able to do so in feedback (see: https://github.com/alphagov/feedback/commit/ebe2a359be75ebf339dafbb405cdbc039cf61218), maybe we can do the same here - depending on which keys and values from the json file we actually need.

This would have the advantage of being able to manage the creds in puppet more easily.

lib/google_analytics/analytics_export_service.rb
```diff
@@ -0,0 +1,31 @@
...
+        json_key_io: File.open(File.join(Rails.root, 'config', 'analytics_exporter_client_secrets.json')),
+        scope: 'https://www.googleapis.com/auth/analytics.edit')
+      @service
+    end
+
+    def export_bad_links(data)
+      File.open('bad_links.csv', 'w') { |file| file.write(data) }
```
h-lame: Maybe we could use a `TempFile` here to avoid colliding if two or more of these run at the same time?

spec/factories/upload_response_factory.rb
```diff
@@ -0,0 +1,13 @@
+module GoogleAnalytics
```
h-lame: Do we need this?  Could it just be an rspec `double`?

lib/tasks/export/analytics_bad_links_exporter.rake
```diff
@@ -0,0 +1,23 @@
+require 'local-links-manager/export/analytics_exporter'
+
+namespace :export do
+  namespace :google_analytics do
+    desc "Export bad links status to Google Analytics"
+    task "bad_links": :environment do
+      service_desc = "Export bad links status to Google Analytics"
+      begin
+        Rails.logger.info("Starting link exporter")
+        Services.icinga_check(service_desc, true, "Starting GA bad link exporter to")
```
h-lame: I feel like the description here is truncated.  "Starting GA bad link exporter to" what?

h-lame (changes requested):
Looks good.  I have a few short comments about the implementation that may or may not be useful.  The major thing is about the JSON file vs. using env vars to achieve the same thing.

elenatanasoiu:
I've added a new commit for updating the `googleauth` gem and then an additional commit to make use of the fixed authentication.

lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
```diff
@@ -14,9 +15,9 @@ def self.local_authority_bad_homepage_url_and_status_csv
         end
       end
 
-      def self.bad_links_url_and_status_csv
+      def self.bad_links_url_and_status_csv(with_ga_headings = false)
```
h-lame: The tests suggest you want this as a keyword arg (you're doing `exporter.bad_links_url_and_status_csv(with_ga_headings: true)`) but this is a normal positional arg.  The reason is works is that the arg in the test is turned into a hash, so `with_ga_headings` ends up being `{with_ga_headings: true}` which _is_ a truthy value, so we get the right output, but for the wrong reasons.

lib/google_analytics/analytics_export_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      @service = AnalyticsService.new
+      @service.authorization ||= ServiceAccountCredentials.make_creds(
+        json_key_io: File.open(File.join(Rails.root, 'config', 'analytics_exporter_client_secrets.json')),
+        scope: 'https://www.googleapis.com/auth/analytics.edit')
+      @service
+    end
+
+    def export_bad_links(data)
+      begin
+        file = Tempfile.new('bad_links.csv')
```
h-lame: I *think* if you use the block form of [`Tempfile.create`](https://ruby-doc.org/stdlib-2.3.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-create) it cleans everything up for you, even if there is an exception.  Worth checking.

lib/tasks/export/analytics_bad_links_exporter.rake
```diff
@@ -7,7 +7,7 @@ namespace :export do
       service_desc = "Export bad links status to Google Analytics"
       begin
         Rails.logger.info("Starting link exporter")
-        Services.icinga_check(service_desc, true, "Exporting bad links list to Google Analytics")
+        Services.icinga_check(service_desc, true, "Exporting list of bad links to Google Analytics")
```
h-lame: Nitpick, but this change could be squashed down into edbd040b364e182878eb3976e56c71142c2d6bdf instead of living in this commit.

docs/google_analytics_custom_import.md
```diff
@@ -0,0 +1,23 @@
+# Authorisation for Google Analytics custom import
```
h-lame, line 1: Nice to see some docs for this :+1:

docs/google_analytics_custom_import.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. Generate a service account (see [Google Documentation](https://developers.google.com/api-client-library/ruby/auth/service-accounts)) and store the JSON key somewhere safe.
+2. Extract the `client_email` value from the JSON key and make it available to the app in the "GOOGLE_CLIENT_EMAIL" environment variable.
+3. Extract the `private_key` value from the JSON key and make it available to the app in the "GOOGLE_PRIVATE_KEY" environment variable. 
+4. Make sure your local_env.yml file contains an entry for: 
+`GOOGLE_ACCOUNT_TYPE: 'service_account'`
```
h-lame: `local_env.yml` is only used in dev mode, but from reading the google lib docs it looks like we'll need the `GOOGLE_ACCOUNT_TYPE` env var set properly for all envs.  What's our plan for production, and should we mention it here?  Maybe we can just assume it's always a 'service_account' and explicitly set the env var to that in code?

h-lame (changes requested):
Previous changes I found are all addressed, but I found more.  Sorry!

docs/google_analytics_custom_import.md
elenatanasoiu: I've set the ENV variable for `GOOGLE_ACOUNT_TYPE` in the code, so it will always be set to `service_account` before we attempt authentication. 

I'm going to add the other five environment variables to `govuk_secrets`:
`GOOGLE_CLIENT_EMAIL`
`GOOGLE_PRIVATE_KEY`
`GOOGLE_EXPORT_ACCOUNT_ID`
`GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID`
`GOOGLE_EXPORT_TRACKER_ID`

lib/google_analytics/analytics_export_service.rb
elenatanasoiu: Yes it does!

lib/local-links-manager/export/bad_links_url_and_status_exporter.rb
elenatanasoiu: Fixed.

docs/google_analytics_custom_import.md
```diff
@@ -18,6 +15,32 @@
...
+
+6. In a development environment, you can test the authentication by adding your environment variables in `local_env.yml`, which should look like this:
+
+---
+GOOGLE_CLIENT_EMAIL: 'linkstatusuploader@badlinksexporter.iam.gserviceaccount.com'
+GOOGLE_PRIVATE_KEY: 'abcd'
+GOOGLE_EXPORT_ACCOUNT_ID: '1234'
+GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID: 'UA-12345-2'
+GOOGLE_EXPORT_TRACKER_ID: 'abcdef'
+
```
h-lame: I think it might be more "correct" to combine 5 & 6 here.  Something like:

> 5.  The app assumes the following environment variables are set and made available to it:
>
>  [LIST OF ENV VARS + DESCRIPTIONS AS IS]
> 
> In development mode you can add these to your `local_env.yml` file and they will be turned into environment variables for you.  In deployed environments we manage these through govuk-puppet and govuk-secrets.

lib/google_analytics/google_analytics_export_credentials.rb
```diff
@@ -0,0 +1,15 @@
...
+module GoogleAnalytics
+  class GoogleAnalyticsExportCredentials
+    def self.authorization(scopes)
+      raise ArgumentError, "Must define GOOGLE_PRIVATE_KEY and GOOGLE_CLIENT_EMAIL in order to authenticate." unless all_configuration_in_env?
+      Google::Auth.get_application_default(scopes)
+    end
+
+    def self.all_configuration_in_env?
+      %w(GOOGLE_PRIVATE_KEY GOOGLE_CLIENT_EMAIL).all? { |env_var| ENV[env_var].present? }
+    end
```
h-lame, line 13: Maybe we should add `GOOGLE_ACCOUNT_TYPE` to these lists?

config/secrets.yml
```diff
@@ -13,13 +13,28 @@
...
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   link_checker_api_secret_token: <%= ENV["LINK_CHECKER_API_SECRET_TOKEN"] %>
+  GOOGLE_CLIENT_EMAIL: 'linkstatusuploader@badlinksexporter.iam.gserviceaccount.com'
+  GOOGLE_PRIVATE_KEY: 'abcdefg'
+  GOOGLE_EXPORT_ACCOUNT_ID: '12345'
+  GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID: 'UA-XXXXX-X'
+  GOOGLE_EXPORT_TRACKER_ID: 'abcdefg'
```
h-lame: Don't think we want these entries in secrets.yml now we're saying we will use env vars.

lib/google_analytics/analytics_export_service.rb
```diff
@@ -18,11 +18,12 @@ def build
...
+        response = ''
         Tempfile.create('bad_links.csv', Dir.pwd) do |file|
           file.write(data)
-          response = service.upload_data(ENV['GA_EXPORT_ACCOUNT_ID'],
-                                         ENV['GA_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID'],
-                                         ENV['GA_EXPORT_TRACKER_ID'],
+          response = service.upload_data(ENV['GOOGLE_EXPORT_ACCOUNT_ID'],
+                                         ENV['GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID'],
+                                         ENV['GOOGLE_EXPORT_TRACKER_ID'],
                                          fields: "accountId,customDataSourceId,errors,id,kind,status,uploadTime",
```
h-lame: This is another git hygiene nitpick, but this change seems unrelated to the rest of this commit, and could happily live in the commit where you introduce this code: e6362afe5100e286606cf6455eea2cf4b51579b9.

h-lame (changes requested):

lib/google_analytics/google_analytics_export_credentials.rb
elenatanasoiu: I've moved the setting of this var inside `GoogleAnalyticsExportCredentials`, so now we always have it set.

h-lame: :+1:

h-lame (approved):
Thanks for dealing with all my pedantry.  Looks good!

PR #136, opened 2017-12-07
Schedule daily upload of bad links to GA

elenatanasoiu:
This will schedule a daily operation to upload a file to Google Analytics. The file will contain a list of broken links and their statuses (reason for being broken). This will allow us to track daily changes to the status of broken links in GA. 

For: https://trello.com/c/9CHX7tje/312-run-daily-scheduled-export-of-bad-links-to-ga

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -22,3 +22,6 @@ every :day, at: '5am' do
   rake 'import:google_analytics'
 end
 
+every :day, at: '6am' do
+  rake 'export:google_analytics:bad_links'
+end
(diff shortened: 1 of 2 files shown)


h-lame (approved):

sihugh:
We might want to consider wrapping the rake task [in a lock](https://github.com/alphagov/local-links-manager/blob/master/lib/tasks/import/google_analytics.rake#L7-L8) like the imports have, otherwise this job will run on each backend server at 6am.  I'm assuming that this would result in duplicate GA content?

Another approach that people have taken to solve the multiple schedules thing is to get Jenkins to run it instead. ([An example](https://github.com/alphagov/govuk-puppet/blob/release/modules/govuk_jenkins/templates/jobs/whitehall_run_broken_link_checker.yaml.erb))

sihugh:
You don't need anything else in order to schedule the job as local links manager is set up to [create schedules on deployment](https://github.com/alphagov/govuk-app-deployment/blob/master/local-links-manager/config/deploy.rb#L11)

elenatanasoiu:
I've added a lock for the export.

sihugh (approved):

sihugh (approved):

PR #162, opened 2018-01-05
Bump govuk_app_config

sihugh:
This enables us to use the same config as the other apps in GOV.UK.

In order to use the logstasher config in the gem we should remove the config
within the app.

We can remove unicorn and logstasher as direct dependencies in the Gemfile as
they are inherited from govuk_app_config too.


cbaines (approved):
Looks good to me 👍

PR #167, opened 2018-01-11
Schedule deleting uploaded files from GA

elenatanasoiu:
For: https://trello.com/c/NYNZzVyW/315-schedule-deleting-files-from-ga

Because we're now uploading a daily list of bad links to GA, we're going to reach the quota limit if we don't start deleting these files automatically. 

This job will run daily, before we upload a new file. For reference and as a sanity check that the daily upload task is still working, we're going to delete all files but the last 2.

Deleting the files after they've been processed does not get rid of the data that was introduced through them. Processing time when a file has been uploaded only takes a few seconds so we are safe if we delete the file 3 days later. 

## Documentation

Documentation for the GA Management API `deleteUploadData` method we're using can be found here: 
- API docs https://developers.google.com/analytics/devguides/config/mgmt/v3/mgmtReference/management/uploads/deleteUploadData
- Ruby wrapper for API
https://github.com/google/google-api-ruby-client/blob/04f0dbfe7aac8229dabaa5ddee32ea03b2402402/generated/google/apis/analytics_v3/service.rb

### DIFF
--- a/lib/google_analytics/analytics_export_service.rb
+++ b/lib/google_analytics/analytics_export_service.rb
@@ -18,6 +18,7 @@ module GoogleAnalytics
 
     def export_bad_links(data)
       begin
+        delete_previous_uploads
         response = ''
         Tempfile.create('bad_links.csv', Dir.pwd) do |file|
           file.write(data)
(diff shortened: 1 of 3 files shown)


lib/google_analytics/analytics_export_service.rb
```diff
@@ -28,8 +29,35 @@ def export_bad_links(data)
...
                                          content_type: 'application/octet-stream')
         end
+        Rails.logger.info "A new file has been uploaded for #{Date.today}"
       end
       response
     end
+
+    def delete_previous_uploads(min_number_to_delete: nil)
+      begin
+        min_number_to_delete = 2 unless min_number_to_delete.present?
```
h-lame: It'd be more idiomatic to use `||=` to set this without the explicit `.present?` check, or we could set a default value in the argument list.

lib/google_analytics/analytics_export_service.rb
```diff
@@ -28,8 +29,35 @@ def export_bad_links(data)
...
+
+        upload_list = service.list_uploads(ENV['GOOGLE_EXPORT_ACCOUNT_ID'],
+                                           ENV['GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID'],
+                                           ENV['GOOGLE_EXPORT_TRACKER_ID'])
+
+        number_of_items = upload_list.items.count
+        return if number_of_items <= min_number_to_delete
+
+        custom_data_import_uids = upload_list.items.last(number_of_items - 2).map(&:id)
```
h-lame: Should the `2` here be `min_number_to_delete` ?

lib/google_analytics/analytics_export_service.rb
```diff
@@ -28,8 +29,35 @@ def export_bad_links(data)
...
+        delete_upload_data_request_object = Google::Apis::AnalyticsV3::DeleteUploadDataRequest.new(custom_data_import_uids: custom_data_import_uids)
+
+        @service.delete_upload_data(
+          ENV['GOOGLE_EXPORT_ACCOUNT_ID'],
+          ENV['GOOGLE_EXPORT_CUSTOM_DATA_IMPORT_SOURCE_ID'],
+          ENV['GOOGLE_EXPORT_TRACKER_ID'],
+          delete_upload_data_request_object
+        )
+
+        Rails.logger.info "Previous uploaded files are now deleted. We've kept the last 2."
```
h-lame: And here.

h-lame (changes requested):
A couple of minor changes, but I also think we should add a test to cover the behaviour of the `delete_previous_uploads` method as it's not trivial when we consider the `min_number_to_delete` thing.

h-lame (approved):

PR #281, opened 2018-08-07
Ensure only strings are sent to icinga check

emmabeynon:
The `code` argument passed to `self.icinga_check` is a boolean value so we
can't call `shellescape` on it. This was causing the `check-links` rake task to
break.

Initially I removed `shellscape` from `code` in the `self.icing_check` method,
however Brakeman is flagging `code` as a potential command injection security
warning because it is a boolean value. The only apparent use of `code` is to be
interpolated in a string which is used to notify Icinga, so it seems reasonable
to send it as a string instead.

This PR also changes other instances of `self.icinga_check` to send `code` as a string.

### DIFF
--- a/lib/tasks/check-links/link_checker.rake
+++ b/lib/tasks/check-links/link_checker.rake
@@ -8,20 +8,20 @@ task "check-links": :environment do
     lock_obtained: ->() {
       begin
         Rails.logger.info("Lock obtained, starting link checker")
-        Services.icinga_check(service_desc, true, "Lock obtained, starting link checker")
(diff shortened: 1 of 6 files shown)


sihugh (approved):
Seems legit.  I was looking at the [send_nsca](http://manpages.ubuntu.com/manpages/trusty/man8/send_nsca.8.html) docs and it suggests that the status should be a numeric return code.  We don't currently do that, and it's a bit outside the scope of this fix, but I wonder if we'd been doing it wrong all this time?

PR #413, opened 2019-04-09
Download all local authority links

AlanGabbianelli:
Allow download of links with specific statuses

The previous behaviour of clicking a link to download broken links has
been modified so that the user can select specific statuses for which
they want to download links. Only links with the selected status will be
present in the downloaded csv file. They are initially all selected by
default.

See specific commit messages for more context on other changes.

Trello: https://trello.com/c/Z6SmqxE3/938-add-button-to-download-all-links-for-a-council

### DIFF
--- a/app/assets/javascripts/local_authorities.js
+++ b/app/assets/javascripts/local_authorities.js
@@ -1,0 +1,17 @@
+document.addEventListener("DOMContentLoaded", function(event) {
+  var checkboxes = document.getElementsByClassName('links_status_checkbox');
+  var links_download_button = document.getElementById('links_download_button');
+  var url = links_download_button.href.split('?')[0];
+
+  for (var i = 0; i < checkboxes.length; i++) {
+    checkboxes[i].addEventListener('change', function() {
+      var params = [];
+      for (var j = 0; j < checkboxes.length; j++) {
+        if (checkboxes[j].checked) {
+          params.push(checkboxes[j].name + '=' + checkboxes[j].value)
+        }
+      }
+      links_download_button.href = url + '?' + params.join('&')
+    });
+  }
+});
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,8 @@ gem 'uglifier', '>= 1.3.0'
 gem 'whenever', require: false
 
 group :development do
+  gem 'better_errors', '~> 2.5.1'
+  gem 'binding_of_caller', '~> 0.8.0'
   gem 'capistrano-rails'
   gem 'web-console', '~> 3.7' # Access an IRB console by using <%= console %> in views
 end
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ group :development, :test do
 end
 
 group :test do
-  gem 'govuk_test'
+  gem 'govuk_test', '~> 0.4.2'
   gem 'capybara', '~> 3.10'
   gem 'timecop'
   gem 'webmock', '~> 3.5.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -54,7 +54,13 @@ GEM
       execjs
     aws-xray-sdk (0.10.2)
       oj (~> 3.0)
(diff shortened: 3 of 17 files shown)


app/assets/javascripts/local_authorities.js
```diff
@@ -0,0 +1,15 @@
...
+
+  links_download_button.addEventListener('mouseenter', function() {
+    var checkboxes = document.getElementsByClassName('link_status_checkbox');
+    var params = [];
+    for (var i = 0; i < checkboxes.length; i++) {
+      if (checkboxes[i].checked) {
+        params.push(checkboxes[i].name + '=' + checkboxes[i].value + '&')
+      }
+    }
+    links_download_button.href = href + '?' + params.join('')
```
thomasleese: I think if we do `params.join('&')` here, we can avoid having to do `+ '&'` in the line above which will lead to there being a trailing `&`.

app/assets/javascripts/local_authorities.js
```diff
@@ -0,0 +1,15 @@
+document.addEventListener("DOMContentLoaded", function(event) {
+  var links_download_button = document.getElementById('links_download_button');
+  var href = links_download_button.href;
+
+  links_download_button.addEventListener('mouseenter', function() {
```
thomasleese: I wonder if it's better to change the href each time a checkbox gets checked/unchecked, because I'm not sure if this event fires say if a user uses a keyboard to navigate to the button.

spec/rails_helper.rb
```diff
@@ -12,6 +12,8 @@
 require 'capybara/rails'
 require 'capybara/rspec/matchers'
 
+Capybara.server = :puma, { Silent: true }
```
thomasleese: I don't think we need this anymore now that we're using `govuk_test`.

AlanGabbianelli: We do if we want to silence Puma logs and prevent it from writing
```
Puma starting in single mode...
* Version 3.10.0 (ruby 2.4.1-p111), codename: Russell's Teapot
* Min threads: 0, max threads: 1
* Environment: test
* Listening on tcp://0.0.0.0:36775
Use Ctrl-C to stop
```
while running the test suite. See https://github.com/rspec/rspec-rails/issues/1897

AlanGabbianelli: ...unless we modify `govuk_test` and put this directly there.

thomasleese: Ah I see! Yeah, I think that sounds like something worth adding to `govuk_test`.

app/assets/javascripts/local_authorities.js
AlanGabbianelli: My initial thought was the same and I started implementing it that way, but talking with @edwardkerry I decided to do it this way so to add it only once instead of once for each checkbox.

That said, I see your point, and I think it a stronger one. I'll amend this now.

edwardkerry: Could we add a second eventListener for `onfocus` to catch that, which I think would capture keyboard navigation? It is a small and quick script but might be good to limit how often it runs if possible

AlanGabbianelli: Thinking about it, I feel that, semantically, it's more accurate to add event listeners to the checkboxes. Checking/unchecking a checkbox (not clicking the button) is what changes the outcome.

edwardkerry: Sounds sensible :+1:

spec/rails_helper.rb
AlanGabbianelli: I'll remove it from here and I'll add it to `govuk_test`

AlanGabbianelli: https://github.com/alphagov/govuk_test/pull/16

AlanGabbianelli: https://github.com/alphagov/govuk_test/pull/17

thomasleese (approved):

PR #420, opened 2019-04-17
Alan+dependabot/bundler/capybara 3.16.2

AlanGabbianelli:


AlanGabbianelli:
Not a solution. I'll revisit once I have a bit more time.

PR #421, opened 2019-04-17
Allow upload new urls for a local authority links

AlanGabbianelli:
As users are now able to download a csv with all links for a specific local authority (see #413), we want to give them the ability to also upload that same csv with new links urls to be able to change in batch as many links urls as they want for that specific local authority.

See specific commit messages for more context on other changes.

Trello: https://trello.com/c/wpk9v5S9/877-allow-content-support-to-change-links-by-uploading-a-csv

Co-Authored-By: Elliot Crosby-McCullough <elliot.cm@gmail.com>

### DIFF
--- a/lib/local-links-manager/import/links.rb
+++ b/lib/local-links-manager/import/links.rb
@@ -1,0 +1,39 @@
+require "csv"
+
+module LocalLinksManager
+  module Import
+    class Links
+      def initialize(local_authority)
+        @local_authority = local_authority
+      end
+
+      def import_links(csv_string)
+        updated = 0
+        CSV.parse(csv_string, headers: true) do |row|
+          next if row["New URL"].blank?
+
+          service = Service.find_by(lgsl_code: row["LGSL"])
+          interaction = Interaction.find_by(lgil_code: row["LGIL"])
+
+          next unless service && interaction
+
+          link = Link.retrieve_or_build(
+            local_authority_slug: local_authority.slug,
+            service_slug: service.slug,
+            interaction_slug: interaction.slug,
+          )
+
+          link.update!(url: row["New URL"])
+
+          updated += 1
+        end
+
+        updated
+      end
+
+    private
+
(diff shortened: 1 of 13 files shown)


elliotcm (commented):
Looks good to me but I was too involved to approve this PR.

lib/local-links-manager/import/links.rb
```diff
@@ -0,0 +1,39 @@
+require "csv"
+
+module LocalLinksManager
```
thomasleese, line 3: I think you can do `class LocalLinksManager::Import::Links` here to avoid the indentation.

elliotcm: You can, but that relies on the modules having been defined first elsewhere so it's usual (though visually annoying) to be explicit about it in each file.

Either way can be linted for but we haven't specified an opinion yet.

https://github.com/alphagov/govuk-lint/blob/3bc900f0b5c30541dd3d467f341a2263728a3ce2/configs/rubocop/other-style.yml#L73-L75

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassAndModuleChildren

thomasleese: Hmm interesting, I've always used the compacted style and haven't had problems with relying on modules having been defined already, but I can see why that might be problematic. I don't really mind either way.

thomasleese (approved):

lib/local-links-manager/import/links.rb
elliotcm: It might be worth testing out and revisiting so we can settle on one or the other.  I'm happy to use the compact style provided it works consistently and doesn't rely on the order of files being loaded.

AlanGabbianelli: Discussed offline: I'll leave it expanded in this PR, and we can revisit once we have an agreed and consistent preferred style.

PR #453, opened 2019-05-31
Various improvements and refactoring

thomasleese:
This PR tidies up a few things in the app to make the code slightly simpler. It auto-requires the `rails_helper` file in the tests, uses the GdsApi service methods and a bundle update.

See commits for individual details.


spec/lib/google_analytics/analytics_import_service_spec.rb
```diff
@@ -1,4 +1,3 @@
-require 'rails_helper'
```
elliotcm, line 1: I'm a bit wary of this change because it prevents people from writing tests which don't load all of rails (which is why `spec_helper` and `rails_helper` are separate files in the first place).  For example, this analytics spec could be changed to `spec_helper` since it doesn't use any Rails stuff.

The only caveats to this are that a) people often use ActiveSupport stuff (like certain extensions of Ruby objects) in their tests even if those tests don't use Rails otherwise, and b) people have a habit of carelessly using `rails_helper` because it requires less thought, and wind up using it anyway.

This change is not out of the question, but it's very much a capitulation to the idea that we don't encourage lightweight tests, and as a testing philosophy leads over time to slower and slower test runs.

Gemfile.lock
```diff
@@ -397,7 +396,7 @@ GEM
       activemodel (>= 5.0)
       bindex (>= 0.4.0)
       railties (>= 5.0)
-    webdrivers (3.8.0)
```
elliotcm, line 400: There's a couple of major version updates in here, are they safe?

lib/local-links-manager/import/local_authorities_importer.rb
```diff
@@ -94,7 +95,8 @@ def mapit_authorities
       end
 
       def mapit_service_response
-        Services.mapit.areas_for_type(self.class.local_authority_types)
+        mapit = GdsApi.mapit(disable_cache: Rails.env.test?)
```
elliotcm: Part of the reason for using a services module is to allow easy swapping out of clients for mock or fake versions in tests.  Adding a direct call here means that less elegant stubbing (or request mocking) would be required to mock the result of this request.

lib/local-links-manager/import/publishing_api_importer.rb
```diff
@@ -76,7 +76,10 @@ def local_transactions
       end
 
       def publishing_api_response
-        Services.publishing_api.get_content_items(document_type: 'local_transaction', per_page: 150)
+        GdsApi.publishing_api_v2.get_content_items(
```
elliotcm: See the other comment regarding the use of a services module.

lib/local-links-manager/import/local_authorities_importer.rb
thomasleese: I don't think using `GdsApi` makes this any harder, you can do `allow(GdsApi).to receive(:mapit).and_return(my_fake_mapit_instance)`.

Gemfile.lock
thomasleese: I tested the branch on integration and the app seems to be working fine, so I think it is all safe. In particular `webdrivers` is only used to run the tests and they're still passing fine.

spec/lib/google_analytics/analytics_import_service_spec.rb
thomasleese: I agree with you about this making tests slower and going against encouraging lightweight tests.

However, your caveats are the reason I still think this is a good idea. At the very least, most (if not all) of the tests are using Rails auto-loading which I think only works if you do `require "rails_helper"`. Plus the fact that all the tests are requiring `rails_helper` anyway means that this specific change doesn't actually make the tests any slower.

lib/local-links-manager/import/local_authorities_importer.rb
elliotcm: It doesn't make it harder but it does make it more magic.  That's what I meant by "less elegant stubbing".

thomasleese: I've changed it now so the `Services` class returns instances from `GdsApi` but the users continue to have `Service.` rather than access `GdsApi` directly.

bevanloon (approved):

PR #477, opened 2019-07-02
Change config for development database

SamJamCul:
This adds a check to see if a docker database is running, and if so
switches across to using its URL instead

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -6,13 +6,16 @@ default: &default
 development:
   <<: *default
   database: local-links-manager_development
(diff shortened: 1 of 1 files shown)


benthorner:
This makes sense - it looks like we have a [different approach](https://github.com/alphagov/content-publisher/blob/master/config/database.yml#L9) elsewhere on GOV.UK though, which seems to be more concise. Can we use that instead?

SamJamCul:
No problem @benthorner, that makes sense! I've made the change now.

benthorner (approved):
👍

PR #545, opened 2019-11-27
Replace govuk-lint with rubocop-govuk and scss_lint-govuk

emmabeynon:
`govuk-lint` has been deprecated so we replace it with `rubocop-govuk` and `scss_lint-govuk` as per the documentation: https://docs.publishing.service.gov.uk/manual/migrate-from-govuk-lint.html


app/models/local_authority.rb
```diff
@@ -7,9 +7,9 @@ class LocalAuthority < ApplicationRecord
       message: "%{value} is not a valid tier",
     }
 
-  has_many :links
-  belongs_to :parent_local_authority, foreign_key: :parent_local_authority_id, class_name: "LocalAuthority"
-  has_many :service_tiers, foreign_key: :tier_id, primary_key: :tier_id
+  has_many :links, dependent: :destroy
```
benthorner, line 10: I wonder if we should just use `restrict_with_error` everywhere, to be on the safe side?

.rubocop.yml
```diff
@@ -6,3 +6,7 @@ Naming/VariableNumber:
     - 'spec/features/links/index_spec.rb'
     - 'spec/requests/api/link_spec.rb'
     - 'spec/features/services/service_show_spec.rb'
+inherit_gem:
+  rubocop-govuk:
+    - config/default.yml
+    - config/rails.yml # add this line for Rails projects
```
benthorner: Do we benefit from the comment here?

spec/models/link_spec.rb
```diff
@@ -203,7 +203,7 @@
 
     it "sets the link status, link warnings and last checked time to an existing homepage url status, warnings and link last checked time" do
       @local_authority = create(:local_authority, status: "broken", link_warnings: @warnings, link_last_checked: "2016-07-14 11:34:09 +0100")
-      @link = create(:link, url: "http://example.com/thing", status: "ok", link_last_checked: Time.now)
+      @link = create(:link, url: "http://example.com/thing", status: "ok", link_last_checked: Time.zone.now, local_authority: @local_authority)
```
benthorner, line 206: This took me a little while to appreciate!

benthorner (commented):
Great stuff! I've got a question about the `dependent: :destroy` change...

app/models/local_authority.rb
emmabeynon: There are some I'm quite confident with e.g. in this case, a Local Authority has many links, and if the Local Authority no longer exists there's no point in keeping its links.  However I'll do another check and change anything where it's less clear-cut.

.rubocop.yml
emmabeynon: Nope, will remove!

spec/models/link_spec.rb
emmabeynon: Yeah I tried explaining in the commit message, but not sure how clear it is (lots of repetition of Local Authority 🙈)

benthorner: That's fine - esoteric factory behaviour is hardly something to fix in this PR :-).

benthorner (approved):
This looks good to me - thanks for working through my comments 😃.

I'm still not entirely comfortable with the use of `dependent: :destroy`, since it represents new (and presumably redundant) functionality: previously, a deletion would not cascade. A safer approach might be to use `restrict` everywhere unless we need to do otherwise - this is the approach we [adopted in content-publisher](https://github.com/alphagov/content-publisher/blob/master/docs/adr/0008-restricted-deletion.md). That said, I can't see this being a problem in practice - people shouldn't be arbitrarily deleting rows in production without investigating the consequences themselves first.

PR #623, opened 2020-04-09
Fix URLs for ESD imports to use HTTPS

boffbowsh:
The import scripts were failing as the ESD Standards site now redirects
from http to https. The scripts don't follow redirects so they were
returning errors. Fixing to use https is preferable to being man-in-
the-middled over http.


issyl0 (approved):
Makes sense to me. Thanks, @boffbowsh!

PR #627, opened 2020-04-15
Enable volunteering service

thomasleese:
We want to start showing local links for this, so we need to enable it in Local Links Manager.

[Trello Card](https://trello.com/c/h89SwLyy/1908-5-add-a-new-covid-19-volunteering-service-to-local-links-manager)


1pretz1 (approved):
I have little to no knowledge of local links manager but am giving it a tick as I've checked through the trello card, code and through previous migrations and it all looks reasonable to me.

PR #631, opened 2020-04-17
Resolve links to parent local authority

thomasleese:
This changes the link resolver class to lookup the links of the parent local authority, if the parent local authority tier is a valid tier for this service.

This is useful because it allows users to be shown a more suitable link if the most local council tier has no link set in the cases where the service is provided by multiple council tiers. This has come up recently as we're adding a new service for volunteering, but don't have all the links available at all tiers.

[Trello Card](https://trello.com/c/ahrs9HWd/1919-make-local-links-manager-fallback)


sihugh (approved):
🍂 🔙 🧝
🎉

boffbowsh:
Nice one 🎉

PR #639, opened 2020-05-01
Add Covid-19 service

huwd:
Trello: [Set up new service in local links manager for local authority covid pages](https://trello.com/c/n5l5Ahwp/356-set-up-new-service-in-llm-for-local-authority-covid-pages)

Add a new service LGSL type for Covid-19 links


brucebolt (approved):
Looks good 👍.  Whilst I still have concerns as to whether this should actually be a migration (as discussed IRL on Friday), it remains consistent with other service code activations, so is probably the best approach to take again.

I've also run this locally and can see service 1287 is now active in the UI and API.

I would perhaps suggest squashing both the commits into one, since the schema version bump and migration are both one functional change.

PR #650, opened 2020-05-17
Run RuboCop as part of `bundle exec rake` defaults

issyl0:
- This is a prerequisite for being able to move more of our apps to
  GitHub Actions. We want one single step bundle exec rake that runs
  everything. This ensures parity between local dev and what CI runs
  (something that we don't always have now).
- Remove old `db/migrate/201*` RuboCop disablement - by default
  it's in the gem now.

https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-task default: [:spec]
+task default: %i[spec lint]
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,5 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-task default: %i[spec lint]
+Rake::Task[:default].clear
+task default: %i[lint spec]
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,10 +7,6 @@ inherit_mode:
   merge:
     - Exclude
 
-AllCops:
-  Exclude:
-    - db/migrate/201*.rb
-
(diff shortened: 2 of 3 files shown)


Rakefile
```diff
@@ -5,4 +5,4 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-task default: [:spec]
+task default: %i[spec lint]
```
benthorner: Same comment about [the behaviour of `spec`](https://github.com/alphagov/content-publisher/blob/master/Rakefile#L7) here.

Are we trying to keep the order of `lint` vs. `spec` consistent? I notice it's different between this and the previous PR I looked at.

issyl0: I think we have bigger fish to fry.

issyl0:
```suggestion
Rake::Task[:default].clear
task default: %i[lint spec]
```

issyl0: But I switched the ordering anyway. 😇

benthorner (approved):
👍

benthorner (approved):

Rakefile
```diff
@@ -5,4 +5,5 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-task default: [:spec]
+Rake::Task[:default].clear
```
benthorner, line 8: Looks like we need `unless Rails.env.production?` here too.

PR #658, opened 2020-05-21
Fix clearing non-existant 'default' task

benthorner:
Clearing the 'default' rake task throws an error in production,
when RSpec hasn't manipulated it into existance.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,5 +5,7 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-Rake::Task[:default].clear
+# RSpec shoves itself into the default task without asking, which confuses the ordering.
+# https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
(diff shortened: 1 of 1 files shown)


barrucadu (approved):

Rakefile
```diff
@@ -5,5 +5,7 @@ require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
 
-Rake::Task[:default].clear
+# RSpec shoves itself into the default task without asking, which confuses the ordering.
+# https://github.com/rspec/rspec-rails/blob/eb3377bca425f0d74b9f510dbb53b2a161080016/lib/rspec/rails/tasks/rspec.rake#L6
+Rake::Task[:default].clear unless Rails.env.production?
```
kevindew, line 10: Rather than doing an environment check can we check the task exists? I believe that'd feel more direct as the production check feels only tangentially related.

benthorner: That would have been a better pattern - unfortunately, this is now in quite a few repos 😬

kevindew: Argh I knew I'd get beaten on the review. We even have a nice method `Rake::Task.task_defined?(:default)`

PR #665, opened 2020-06-03
Stop links with 0 clicks appearing at the top of the list.

callumknights:
Made a small change to `anaytics_importer.rb` to prevent the number of clicks
for a given link ever being nil, as when GA shows no clicks for a link it gives
that as `nil`, this breaks the ordering for the links.

The data migration will be the first of two, there will be a second migration adding a constraint to the `analytics` column, but's it's safer to add this separately so that will come in a follow up PR.

### DIFF
--- a/lib/local-links-manager/import/analytics_importer.rb
+++ b/lib/local-links-manager/import/analytics_importer.rb
@@ -26,7 +26,7 @@ module LocalLinksManager
         link = Link.lookup_by_base_path(item[:base_path])
 
         if link
-          link.update!(analytics: item[:clicks])
+          link.update!(analytics: item[:clicks] || 0)
           summariser.increment_updated_record_count
           @processed_ids.add(link.id)
         else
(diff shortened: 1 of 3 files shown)


Tetrino:
Have you talked to data to see if this would break any of their tools?

lib/local-links-manager/import/analytics_importer.rb
```diff
@@ -26,7 +26,7 @@ def import_item(item, _response, summariser)
         link = Link.lookup_by_base_path(item[:base_path])
 
         if link
-          link.update!(analytics: item[:clicks])
+          link.update!(analytics: item[:clicks] || 0)
```
thomasleese, line 29: Just had a thought that it might be nicer to do this:

```suggestion
          link.update!(analytics: item.fetch(:clicks, 0))
```

But this depends on whether `:clicks` exists in `item` but is `nil`, or if it doesn't exist in `item`. The solution you've already got using `||` handles both cases, so perhaps it's okay to stick with this one.

thomasleese (approved):

lib/local-links-manager/import/analytics_importer.rb
callumknights: I'd personally feel it's better to leave both cases handled, given I'm still not 100% how GA handles this figure.

PR #682, opened 2020-07-01
Fix CSV upload where row contains invalid URL

ChrisBAshton:
This PR makes the 'CSV import links' feature more robust, so that if it encounters a row which has a malformed URL it will fail gracefully rather than throw the default Rails 'something went wrong' page.

I've customised the show.html.erb file a little to allow for array-like errors, separated by `<br>`.

It also includes some README improvements, as getting the app set up took a little while.

|Before|After|
|------|-----|
||Screenshot 2020-07-03 at 10 18 47|

Trello: https://trello.com/c/5IFlmQQQ/2039-3-improve-local-links-manager-error-handling-for-the-csv-upload-feature


db/seeds.rb
```diff
@@ -5,3 +5,5 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+User.create!(name: "Foo", email: "foo@gov.uk")
```
thomasleese: I think it's better to use `find_or_create!` here so that the seed task copes with existing data and can be run multiple times.

thomasleese (approved):
This looks good to me - I've left one comment but I don't feel particularly strongly about it.

lib/local-links-manager/import/links.rb
```diff
@@ -1,31 +1,47 @@
...
+          link = Link.retrieve_or_build(slugs)
+
+          begin
+            link.update!(url: new_url)
+            updated += 1
+          rescue ActiveRecord::RecordInvalid => e
+            GovukError.notify(
+              LocalLinksManager::Import::UrlValidationException.new(e.message),
+              extra: slugs.merge(link_id: link.id),
+            )
```
kevindew, line 41: Sorry for a drive-by comment.

We shouldn't use Sentry for logging validation errors in a users input as per [rfc-87](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md#2-notifications-should-be-actionable): 

```
### 2. Notifications should be actionable

Sentry notifications should be something that requires a developer of the app to do something about it. Not just a piece of information.
```

As these create error logs that aren't actionable or resolvable and may appear at any cadence.

The approach we've used to record information where we want a record of user errors is to write to the Rails logger so that the issue is available in Kibana.

ChrisBAshton: Thanks for the comment, Kevin. Sentry logging was a request on the Trello card, but in hindsight it does feel a bit odd to be logging an error the developer shouldn't do anything about.

I'll talk to Platform Health about a follow-up piece of work to swap this out for Rails logging.

PR #695, opened 2020-08-04
Make app consistent with upgrades

cdccollins:
This makes local-links-manager consistent with the other apps that have been upgraded to Rails 6


EGiataganas:
Hey @cdccollins this looks great, I've compared the pr changes with [railsdiff][1] and there are some changes. Can we do the following changes :

1. Remove `bin/bundle`, `bin/update`
2. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
3. In the `config/puma.rb` Rails 6 introduce `has max_threads_count` and `min_threads_count`. It will be good to update it
4. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2

Rhian:
Here are a couple of tiny optional updates that could also be done:
- Delete the empty `app/assets/javascripts/application.js ` file
- Explicitly add the local dev link `http://local-links-manager.dev.gov.uk/` to config hosts in development.rb, as per https://github.com/alphagov/travel-advice-publisher/pull/919/commits/2689aa4690a098a805bc23aebec8aa7d9c1e3723
- Remove the unused (commented) Railties require statements in `/config/application.rb`
- Remove `config/initializers/new_framework_defaults.rb` as per the reason in https://github.com/alphagov/travel-advice-publisher/pull/919/commits/8b0745e9374503aeb2e11511aa0e9b5e0ce5752f
- Add the Rails version number to the top 8 migrations [5.0] as otherwise a new migration will fail

cdccollins:
> * Remove `config/initializers/new_framework_defaults.rb` as per the reason in [alphagov/travel-advice-publisher@8b0745e](https://github.com/alphagov/travel-advice-publisher/commit/8b0745e9374503aeb2e11511aa0e9b5e0ce5752f)

If I remove that file, everything breaks... Is it best then to
1. Leave that file there
2. Move the config in there to application.rb

Rhian:
I'd suggest the first option:

 1. Leave that file there

Rhian:
All changes that were requested look fine.

Rhian (approved):

PR #699, opened 2020-08-28
Fix incorrect icinga check names

kevindew:
In 35414db8 there was a mass find and replace that changed all instances
of local-links-manager to local_links_manager, this unfortunately had
the consequence that any icinga check communication was being made to a
non-existent check. This reverts the naming to local-links-manager.

/cc @cdccollins just so you're aware.


thomasleese (approved):

sophiebrown01 (approved):
Nice catch 💯

PR #722, opened 2020-11-05
Enable Test and Trace Support Payment service

gclssvglx:
We want to start showing local links for this, so we need to enable it
in Local Links Manager.

[Trello](https://trello.com/c/gfu0Sp2q/608-set-up-a-new-service-in-local-links-manager)

### DIFF
--- a/db/migrate/20201105085759_enable_service1826.rb
+++ b/db/migrate/20201105085759_enable_service1826.rb
@@ -1,0 +1,17 @@
+class EnableService1826 < ActiveRecord::Migration[6.0]
+  def up
+    service = Service.find_by(lgsl_code: 1826)
+    interaction = Interaction.find_by(lgil_code: 8)
+
+    service.update!(enabled: true)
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM 👍 
I wonder if it's worth creating a Rake task so that we can stop writing [migrations like these](https://github.com/alphagov/local-links-manager/blob/37b9d6aa99da590d3e2db54c375e0714aad326d5/db/migrate/20200501124957_enable_service_1287.rb)?

gclssvglx:
> LGTM 👍
> I wonder if it's worth creating a Rake task so that we can stop writing [migrations like these](https://github.com/alphagov/local-links-manager/blob/37b9d6aa99da590d3e2db54c375e0714aad326d5/db/migrate/20200501124957_enable_service_1287.rb)?

Great suggestion @ChrisBAshton - it doesn't seem sensible to create a migration just to update some data 👍

PR #737, opened 2020-12-23
Add country name to local authority

gclssvglx:
## What

The country name - which is available from the [Mapit](https://github.com/alphagov/mapit) API - is not present in the current response.

## Why

So that we can provide a better user experience and use the country name where appropriate.

[Trello](https://trello.com/c/QSY1CoAB/643-test-and-trace-nation-validation)


theseanything:
Superseded by https://github.com/alphagov/local-links-manager/pull/738

PR #741, opened 2021-01-04
Add new test to increase coverage for continuous deployment

callumknights:
Added a new test that tests the 'destroy' path to delete a given link.

Also removed some redundant code that would previously cause the 'destroy'
method to throw an error and redirect back if an invalid link was given.
Such scenarios are now automatically caught by the ActiveRecord validations
since this change (https://github.com/alphagov/local-links-manager/commit/ac607b4888fecf55b741099428ef18f06e08fc61)

https://trello.com/c/uAx6eKAQ/2276-activate-continuous-deployment-for-local-links-manager

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -35,9 +35,6 @@ class LinksController < ApplicationController
   def destroy
     if @link.make_missing
       redirect("deleted")
-    else
-      flash[:danger] = "Could not delete link."
-      redirect_back
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/links_controller.rb
```diff
@@ -35,9 +35,6 @@ def update
   def destroy
     if @link.make_missing
```
thomasleese: Since this calls `save!` underneath, we don't need the `if` anymore either.

callumknights: Ahh yes, good catch 👍

thomasleese (approved):

PR #776, opened 2021-03-17
Add functionality to change the LGSL code of a service

gclssvglx:
## What

This adds the functionality to change the LGSL code of a service in `LocalLinksManager`.

To change the LGSL code of a service fully it involves at least two systems: `LocalLinksManager` and [`Publisher`](https://github.com/alphagov/publisher).  It may also include [`Frontend`](https://github.com/alphagov/frontend) if the service is a `LocalTransaction` that is not available in one or more of the devolved nations.  

The outline of the overall process is:

1. [OPTIONAL STEP] In `Frontend` - duplicate any configuration that may be required by the service in [`local_transaction_config.yml`](https://github.com/alphagov/frontend/blob/master/lib/config/local_transaction_config.yml) 

2. In `LocalLinksManager` - create a [duplicate](https://github.com/alphagov/local-links-manager/pull/776/commits/c4c3aa37faa3542e1b0b01944b9789b34d478f13) of the service and all its child records.  The child records need to be linked to the new service.  As both the `label` and `slug` of a service must be unique - the duplicate is created with using a temporary `label` and `slug` that is based on the actual one in the old service but different enough to allow the duplication to take place.

3. In `Publisher` - update the LGSL codes so that the equivalent service there points to the new service in `LocalLinksManager`.

4. In `LocalLinksManager` - [remove](https://github.com/alphagov/local-links-manager/pull/776/commits/28edb6aad7f370339f9ff96cbe528b6b5bfada01) the old service and all its child records.  This switches the service over from the old service to the new (duplicated) service.

5. In `LocalLinksManager` - [update](https://github.com/alphagov/local-links-manager/pull/776/commits/0eb817141f1cb7451e73cdae88700a7b62fe3dbf) the service `label` and `slug` to correct the temporary versions given to them in the duplication step above.

6. In `Publisher` - update the service LGSL code in [local_services.csv](https://github.com/alphagov/publisher/blob/master/data/local_services.csv) from the old to the new service.  This is done because if the [`local_transactions:fetch_and_clean`](https://github.com/alphagov/publisher/blob/master/lib/tasks/local_transactions.rake) rake task were to be re-run - the old service would be re-created at that point and the new service would not - leading to an un-desirable state.

7.  [OPTIONAL STEP] In `Frontend` - remove any configuration that may exist for the old service in [`local_transaction_config.yml`](https://github.com/alphagov/frontend/blob/master/lib/config/local_transaction_config.yml) 

## Why

When the [`find-covid-19-lateral-flow-test-site`](https://www.gov.uk/find-covid-19-lateral-flow-test-site) service was originally created, it used a dummy LGSL code (100001).  We now have a real LGSL code for this service (1827) and would like to use this LGSL code instead of the dummy code.

[Trello](https://trello.com/c/MRGj9keG/210-local-lookup-lateral-mass-test-apply-the-proper-assigned-service-code-1827-agreed-with-local-government-authority)

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -4,7 +4,7 @@ class Service < ApplicationRecord
   has_many :service_interactions, dependent: :destroy
   has_many :links, through: :service_interactions
   has_many :interactions, through: :service_interactions
-  has_many :service_tiers, dependent: :restrict_with_error
+  has_many :service_tiers, dependent: :destroy
   has_many :local_authorities, through: :service_tiers
 
   scope :enabled, -> { where(enabled: true) }
--- a/lib/tasks/service.rake
+++ b/lib/tasks/service.rake
@@ -36,4 +36,26 @@ namespace :service do
 
     service_interaction.update!(live: true)
   end
+
+  desc "Duplicates an existing Service"
+  task :duplicate, %w[from_lgsl_code to_lgsl_code] => :environment do |_, args|
+    old_service = Service.find_by!(lgsl_code: args.from_lgsl_code)
+
+    ActiveRecord::Base.transaction do
+      new_service = old_service.dup
+
+      new_service.assign_attributes(
+        lgsl_code: args.to_lgsl_code,
+        label: "Transitioning #{old_service.label}",
+        slug: "transitioning-#{old_service.slug}",
+      )
+      new_service.save!
+
+      new_service.service_interactions = old_service.service_interactions.map do |si|
+        si.dup.tap { |new_si| new_si.links = si.links.map(&:dup) }
+      end
+
+      new_service.service_tiers = old_service.service_tiers.map(&:dup)
+    end
+  end
 end
--- a/lib/tasks/service.rake
+++ b/lib/tasks/service.rake
@@ -64,4 +64,10 @@ namespace :service do
     service = Service.find_by!(lgsl_code: args.lgsl_code)
     service.update!(label: args.label, slug: args.slug)
   end
+
+  desc "Destroys an existing Service and all dependant records"
+  task :destroy, %w[lgsl_code] => :environment do |_, args|
+    service = Service.find_by!(lgsl_code: args.lgsl_code)
+    service.destroy!
+  end
 end
--- a/lib/tasks/service.rake
+++ b/lib/tasks/service.rake
@@ -58,4 +58,10 @@ namespace :service do
       new_service.service_tiers = old_service.service_tiers.map(&:dup)
     end
   end
+
+  desc "Updates the label and slug of an existing Service"
+  task :rename, %w[lgsl_code label slug] => :environment do |_, args|
+    service = Service.find_by!(lgsl_code: args.lgsl_code)
+    service.update!(label: args.label, slug: args.slug)
+  end
 end
--- a/spec/lib/tasks/service_spec.rb
+++ b/spec/lib/tasks/service_spec.rb
@@ -1,0 +1,25 @@
+describe "Service tasks" do
+  describe "service:duplicate" do
+    it "duplicates the service" do
+      old_service = create(:service, :all_tiers, lgsl_code: 1)
+
+      service_interaction1 = create(:service_interaction, service: old_service)
+      create_list(:link, 3, service_interaction: service_interaction1)
+
+      service_interaction2 = create(:service_interaction, service: old_service)
+      create_list(:link, 3, service_interaction: service_interaction2)
+
+      args = Rake::TaskArguments.new(%i[from_lgsl_code to_lgsl_code], [1, 2])
+      Rake::Task["service:duplicate"].execute(args)
+
+      new_service = Service.find_by!(lgsl_code: 2)
+
+      expect(new_service.label).to eq("Transitioning #{old_service.label}")
+      expect(new_service.slug).to eq("transitioning-#{old_service.slug}")
+
+      expect(new_service.interactions.count).to eq(old_service.interactions.count)
+      expect(new_service.service_interactions.count).to eq(old_service.service_interactions.count)
+      expect(new_service.links.count).to eq(old_service.links.count)
+    end
+  end
+end
--- a/spec/lib/tasks/service_spec.rb
+++ b/spec/lib/tasks/service_spec.rb
@@ -33,4 +33,18 @@ describe "Service tasks" do
         .and change { service.reload.slug }.from(service.slug).to("updated-slug")
     end
   end
+
+  describe "service:destroy" do
+    it "should destroy the service" do
+      service = create(:service, :all_tiers, lgsl_code: 1)
+      service_interaction = create(:service_interaction, service: service)
+      create_list(:link, 3, service_interaction: service_interaction)
+      args = Rake::TaskArguments.new(%i[lgsl_code], [1])
+
+      expect { Rake::Task["service:destroy"].execute(args) }
+        .to change { Service.count }.from(1).to(0)
+        .and change { ServiceInteraction.count }.from(1).to(0)
+        .and change { Link.count }.from(3).to(0)
+    end
+  end
 end
--- a/spec/lib/tasks/service_spec.rb
+++ b/spec/lib/tasks/service_spec.rb
@@ -22,4 +22,15 @@ describe "Service tasks" do
       expect(new_service.links.count).to eq(old_service.links.count)
     end
   end
+
+  describe "service:rename" do
+    it "should update the label and the slug" do
+      service = create(:service, :all_tiers, lgsl_code: 1)
+      args = Rake::TaskArguments.new(%i[lgsl_code label slug], [1, "Updated label", "updated-slug"])
+
+      expect { Rake::Task["service:rename"].execute(args) }
+        .to change { service.reload.label }.from(service.label).to("Updated label")
+        .and change { service.reload.slug }.from(service.slug).to("updated-slug")
+    end
+  end
 end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -32,6 +32,8 @@ Dir[Rails.root.join("spec/support/**/*.rb")].sort.each { |f| require f }
 # If you are not using ActiveRecord, you can remove this line.
 ActiveRecord::Migration.maintain_test_schema!
 
+Rails.application.load_tasks
+
 RSpec.configure do |config|
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
   config.fixture_path = "#{::Rails.root}/spec/fixtures"


app/models/service.rb
```diff
@@ -46,4 +46,40 @@ def required_tiers(tier_name)
...
+
+      service_interaction.links.each do |link|
+        new_link = link.dup
+        new_link.service_interaction_id = new_service_interaction.id
+        new_link.save!
+      end
+    end
+  end
+
+  def remove!
```
kevindew: This method seems to be a synonym of delete or destroy which are methods this already has. 

I feel like we could use the existing `destroy!` method rather than needing an additional method for this. It looks like service_interaction and links will be destroyed automatically. However it looks like service_tiers may [block it](https://github.com/alphagov/local-links-manager/blob/7e780dd1b5dc058ecfd4439e834dd0af60ba4cb9/app/models/service.rb#L7), looking at the commit message I think it'd be ok for us to switch that to be a dependent destroy.

lib/tasks/service.rake
```diff
@@ -36,4 +36,22 @@ namespace :service do
...
+  end
+
+  desc "Updates the label and slug of an existing Service"
+  task :rename, %w[lgsl_code label slug] => :environment do |_, args|
+    service = Service.find_by(lgsl_code: args.lgsl_code)
+    service.update!(label: args.label, slug: args.slug)
+  end
+
+  desc "Removes an existing Service and all dependant records"
+  task :remove, %w[lgsl_code] => :environment do |_, args|
```
kevindew: Since we're planning to keep this task it'd be good to have a unit test for it.

I also feel like, if we switch remove! to destroy! on the service that we should name this task as destroy to match what Rails does.

app/models/service.rb
```diff
@@ -46,4 +46,40 @@ def required_tiers(tier_name)
   def to_param
     slug
   end
+
+  def duplicate!(new_lgsl_code)
+    new_service = dup
+    new_service.lgsl_code = new_lgsl_code
+    new_service.label = "TRANSITION -- #{label}"
+    new_service.slug = "TRANSITION -- #{slug}"
```
kevindew: This doesn't seem to be formatted as a slug anymore (I'd not expect spaces or capitals) more `transition-`

I imagine for both of these they'd benefit from being a verb like transitioning to indicate that something is happening rather than being static.

app/models/service.rb
```diff
@@ -46,4 +46,40 @@ def required_tiers(tier_name)
   def to_param
     slug
   end
+
+  def duplicate!(new_lgsl_code)
```
kevindew: Since this is just temporary code I'd suggest having it entirely within the rake task so the model doesn't pick up any churn from it. Apologies if that conflicts with suggestions Leena made on similar code but I think it makes sense if we're going to run and then delete to reduce the files that are touched.

By the way this method shouldn't have a bang suffix as there isn't a non bang alternative: https://rubystyle.guide/#dangerous-method-bang

app/models/service.rb
```diff
@@ -46,4 +46,40 @@ def required_tiers(tier_name)
...
+    new_service = dup
```
kevindew: Nice use of dup 👍

app/models/service.rb
```diff
@@ -46,4 +46,40 @@ def required_tiers(tier_name)
...
+      new_service_interaction = service_interaction.dup
+      new_service_interaction.service_id = new_service.id
+      new_service_interaction.save!
+
+      service_interaction.links.each do |link|
+        new_link = link.dup
+        new_link.service_interaction_id = new_service_interaction.id
+        new_link.save!
+      end
+    end
```
kevindew: I'm being nit-picky for temporary code but you can make this code a bit shorter by relying on relationship autosaving:

```
  new_service = dup
  new_service.assign_attributes(
    lgsl_code: lgsl_code,
    label: "TRANSITION -- #{label}",
    slug: "TRANSITION -- #{slug}",
  )
  new_service.save!
  
  new_service.service_interactions = service_interactions.map do |si|
    si.dup.tap { |new_si| new_si.links = si.links.map(&:dup) }
  end
  new_service.service_tiers = service_tiers.map(&:dup)
```

kevindew: Oh and I also think we should also put it in a transaction so it's all or nothing if that are any problems.

spec/models/service_spec.rb
```diff
@@ -53,4 +53,52 @@
...
+
+  describe "#duplicate!" do
+    it "duplicates the service" do
+      service = create(:service, :all_tiers, lgsl_code: 1)
+
+      interaction1 = create(:interaction, lgil_code: 1, label: "interaction1")
+      service_interaction1 = create(:service_interaction, service: service, interaction: interaction1)
+      create(:link, service_interaction: service_interaction1, url: "https://example.com")
+      create(:link, service_interaction: service_interaction1, url: "https://example.com/test")
+      create(:link, service_interaction: service_interaction1, url: "https://example.com/test/again")
```
kevindew: I don't think you need the URL's on these. You can shorten it to `create_list(:link, 3 service_interaction: service_interaction1)`

spec/models/service_spec.rb
```diff
@@ -53,4 +53,52 @@
       expect(service.required_tiers("all")).to eq([2, 3, 1])
     end
   end
+
+  describe "#duplicate!" do
+    it "duplicates the service" do
+      service = create(:service, :all_tiers, lgsl_code: 1)
+
+      interaction1 = create(:interaction, lgil_code: 1, label: "interaction1")
```
kevindew: I don't think you need to create this, one would be implicitly created when you create a service_interaction

spec/models/service_spec.rb
```diff
@@ -53,4 +53,52 @@
...
+      service.duplicate!(2)
+
+      new_service = Service.find_by(lgsl_code: 2)
+
+      expect(new_service.label).to eq("TRANSITION -- #{service.label}")
+      expect(new_service.slug).to eq("TRANSITION -- #{service.slug}")
+
+      expect(service.interactions.count).to eq(2)
+      expect(service.service_interactions.count).to eq(2)
+      expect(service.links.count).to eq(6)
```
kevindew: These assertions seem a bit strange as they're essentially testing the factory works and not the method under test. It'd seem to make more sense to test: `expect(new_service.interactions.count).to eq(service.interactions.count)`

kevindew (commented):
Great - this looks like it'll do the trick. I've put a few comments on suggestions to improve the approach.

Could you also put some descriptions in your commit messages about why we're doing this.

app/models/service.rb
gclssvglx: Thanks - fixed up as per your suggestion

lib/tasks/service.rake
gclssvglx: Thanks - fixed up as per your suggestion

app/models/service.rb
gclssvglx: Thanks - fixed up as per your suggestion

app/models/service.rb
gclssvglx: Thanks - fixed up as per your suggestion

app/models/service.rb
gclssvglx: Thanks - fixed up as per your suggestion

spec/models/service_spec.rb
gclssvglx: Thanks - fixed up as per your suggestion

spec/models/service_spec.rb
gclssvglx: Thanks - fixed up as per your suggestion

spec/models/service_spec.rb
gclssvglx: Thanks - fixed up as per your suggestion

app/models/service.rb
gclssvglx: Thanks!

spec/models/service_spec.rb
```diff
@@ -53,4 +53,19 @@
...
+
+      service_interaction = create(:service_interaction, service: service)
+      create_list(:link, 3, service_interaction: service_interaction)
+
+      service.destroy!
+
+      expect(service.service_interactions.count).to eq(0)
+      expect(service.links.count).to eq(0)
+      expect(Service.find_by(lgsl_code: 1)).to be_nil
+    end
```
kevindew: I don't think we need this unit test as it's not a method we've wrote - we're essentially unit testing Rails here.

I think we should have a very brief test for the rake task though - mostly for the coverage.

spec/lib/local-links-manager/service_spec.rb
```diff
@@ -0,0 +1,25 @@
+describe "service" do
+  context ":duplicate" do
```
kevindew:
```suggestion
describe "Service tasks" do
  describe "service:duplicate" do
```

This file also seems to be in an unconventional place, a more conventional one would be spec/lib/tasks/service_spec.rb

spec/lib/local-links-manager/service_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      Rake::Task["service:duplicate"].execute(args)
+
+      new_service = Service.find_by!(lgsl_code: 2)
+
+      expect(new_service.label).to eq("Transitioning #{old_service.label}")
+      expect(new_service.slug).to eq("transitioning-#{old_service.slug}")
+
+      expect(new_service.interactions.count).to eq(old_service.interactions.count)
+      expect(new_service.service_interactions.count).to eq(old_service.service_interactions.count)
+      expect(new_service.links.count).to eq(old_service.links.count)
```
kevindew: This looks good.

spec/rails_helper.rb
```diff
@@ -32,6 +32,8 @@
 # If you are not using ActiveRecord, you can remove this line.
 ActiveRecord::Migration.maintain_test_schema!
 
+Rails.application.load_tasks
```
kevindew, line 35: 👍

kevindew (commented):
Just a couple of testing issues and this looks good to merge

kevindew:
Oh I just noticed the commit messages, you shouldn't be having the same description in every message to describe the why. They end up not making a lot of sense. For example:

```
Add functionality to destroy a service …

Why

When the find-covid-19-lateral-flow-test-site service was originally created, it used a dummy LGSL code (100001). We now have a real LGSL code for this service (1827) and would like to use this LGSL code instead of the dummy code.
```

This doesn't really tell us at all why we've added a rake task to destroy a service, which is presumably due to the fact that this can't be done through the user interface and that we have a need to do it because we created a temporary service. We'd also want to explain why we've changed the `has_many :service_tiers, dependent: :restrict_with_error` and no longer feel that it needs to restrict with an error. 

You may find having a read of https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages to help you understand organisational expectations of commit messages, maybe check out some of Leena's commit's for examples they're good examples: https://github.com/alphagov/collections-publisher/commit/7787bb7360bbe13cfce8fc9fa37f1c200cbafc54

spec/lib/tasks/service_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it "should rename the service label and slug" do
+      create(:service, :all_tiers, lgsl_code: 1)
+
+      allow_any_instance_of(Service).to receive(:update!)
+        .with({ label: "Updated label", slug: "updated-slug" })
+        .and_return(true)
+
+      args = Rake::TaskArguments.new(%i[lgsl_code label slug], [1, "Updated label", "updated-slug"])
+      Rake::Task["service:rename"].execute(args)
+    end
```
kevindew, line 34: This test doesn't have any expectations and so doesn't actually test anything.

I'd suggest changing it round to something like:

```
service = create(:service, :all_tiers, lgsl_code: 1)
args = Rake::TaskArguments.new(%i[lgsl_code label slug], [1, "Updated label", "updated-slug"])
expect { Rake::Task["service:rename"].execute(args) }
  .to change { service.reload.label }
  .from(
```
etc

spec/lib/tasks/service_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  describe "service:destroy" do
+    it "should destroy the service" do
+      create(:service, :all_tiers, lgsl_code: 1)
+
+      allow_any_instance_of(Service).to receive(:destroy!).and_return(true)
+
+      args = Rake::TaskArguments.new(%i[lgsl_code], [1])
+      Rake::Task["service:destroy"].execute(args)
+    end
+  end
```
kevindew, line 49: Similar point about this not having any expectations as the above.

kevindew (commented):
Thanks - the commit messages look much improved. Some issues with the tests still to resolve.

kevindew (approved):
Great - thanks for getting this sorted.

PR #789, opened 2021-04-21
Consolidate tasks to import service links from CSV

Rosa-Fox:
[Trello](https://trello.com/c/OTMnFRap/328-spike-dfe-local-consider-import-strategy-for-their-initial-links-and-determining-if-any-further-dev-work-is-needed)

What?
- The two deleted tasks were very similar. The difference was that links.rake
found a LocalAuthority based on its 'name' property and covid19_test_and_trace_scheme_links.rake
found a LA based on its 'slug' property.
- Finding by 'slug' feels more reliable as a slug is less likely to have spaces or characters
that might cause problems in the CSV data.

The encoding has been set to `encoding: 'bom|utf-8'`.

The reason for this is explained [in this SO post](https://stackoverflow.com/a/50336647). Someone "discovered that saving from an .xlsx workbook in MS Excel, to the .csv is what caused those strange characters to appear". I did the same thing and ended up getting confused when it wasn't working due to bytes being added which are unseen to the human eye... this was fixed when re-writing the data in a text editor, but it might save someone else some pain.

These characters were added is a result of [Byte Order Mark (BOM) for UTF-8 encoding](https://estl.tech/of-ruby-and-hidden-csv-characters-ef482c679b35).

To run the task (stolen from Graham's original commit):

The rake task takes 3 params: LGSL code, LGIL code and a filename. The
file to which the filename points to must be a CSV file with 2 columns.
It must have a header row of "slug" and "url". It is expected that this
file be available locally, in the /tmp directory for example.


lib/tasks/import/service_links_from_csv.rake
```diff
@@ -14,8 +14,8 @@ namespace :import do
     imported = 0
 
     csv.each do |row|
-      slug = row["slug"]&.strip
-      url = row["url"]&.strip
+      slug = row[0]&.strip
+      url = row[1]&.strip
```
gclssvglx: I feel using column position is somewhat risky - far too easy to switch columns around.  Using headers seems much safer.  

Microsoft Excel puts/keeps this byte string at the start of every line `\xEF\xBB\xBF` so I'm concerned that we are storing those characters, as they are not being removed from the actual column value.  Which may have undesirable consequences and tricky to track bugs.  Thankfully, Google Sheets doesn't seem to do this!  

We could strip this stuff out `.gsub("\xEF\xBB\xBF".force_encoding("UTF-8"), "")` for example.  But I wonder if we could just 'encourage' folks to use Google Sheets to do the conversion?

gclssvglx (commented):
These are nice changes @Rosa-Fox - makes the whole thing more general and readable 👍

lib/tasks/import/service_links_from_csv.rake
gclssvglx: Looking into this a bit further, it seems that only the first row contains `\xEF\xBB\xBF` and not all rows.  So, there should be no issues with saving this stuff.  Assuming we use headers.

gclssvglx: @Rosa-Fox I'll approve this and leave the choice of `row[0]` or `row["slug"]` to you.

gclssvglx (approved):
Nice!

Rosa-Fox:
Thanks @gclssvglx. I tried to gsub but kept getting an error about modifying a frozen string. I changed it to `encoding: 'bom|utf-8'` encoding instead which seems work for both and keep the data in tact. I also added some docs. Let me know what you think :)

gclssvglx:
Love the docs @Rosa-Fox nice touch to add those 👍

PR #850, opened 2021-11-11
Add Local Custodian Codes

baisa:

Added Rake task responsible for populating local_custodian_code column with data from the CSV file. `import_from_csv` method is implemented in `local_custodian_codes_importer` helper.

[Trello card](https://trello.com/c/OQKm9z4V/2758-add-local-custodian-codes-to-local-links-manager-5)


spec/lib/local-links-manager/import/local_custodian_codes_importer_spec.rb
```diff
@@ -0,0 +1,42 @@
+describe LocalLinksManager::Import::LocalCustodianCodesImporter do
+  describe "#import_from_csv" do
+    context "local custodian codes for Local Authorities in Great Britain" do
+      let!(:local_authority_1) { create(:local_authority, slug: "leicestershire", local_custodian_code: nil) }
+      let!(:local_authority_2) { create(:local_authority, slug: "surrey-heath", local_custodian_code: nil) }
+      let!(:local_authority_3) { create(:local_authority, slug: "stockton-on-tees", local_custodian_code: nil) }
+
+      it "populates code with value in different formats" do
+        LocalLinksManager::Import::LocalCustodianCodesImporter.new.import_from_csv("spec/lib/local-links-manager/import/fixtures/addressbase-local-custodian-codes-test.csv")
```
ChrisBAshton: This can be refactored slightly:

```suggestion
        described_class.new.import_from_csv("spec/lib/local-links-manager/import/fixtures/addressbase-local-custodian-codes-test.csv")
```

spec/lib/local-links-manager/import/local_custodian_codes_importer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(local_authority_1.local_custodian_code).to eq "2460"
+        expect(local_authority_2.local_custodian_code).to eq "3640"
+        expect(local_authority_3.local_custodian_code).to eq "738"
+      end
+    end
+
+    context "local custodian codes for Local Authorities in Northern Ireland" do
+      let!(:local_authority_1) { create(:local_authority, slug: "belfast", local_custodian_code: nil) }
+
+      it "populates code with value" do
```
ChrisBAshton: As a general rule, we want to avoid using `context` when it only wraps around a single test. It's a sign that it's been slightly over-engineered. These tests would read more clearly as something like:

```ruby
it "extracts local custodian codes from CSVs containing Northern Ireland Local Authorities" do
  local_authority = create(:local_authority, slug: "belfast", local_custodian_code: nil)
  described_class.new.import_from_csv("spec/lib/local-links-manager/import/fixtures/addressbase-local-custodian-code-Northern-Ireland-test.csv")

  expect(local_authority_1.reload.local_custodian_code).to eq "8132"
end
```

spec/lib/tasks/local_custodian_codes_from_csv_spec.rb
```diff
@@ -0,0 +1,10 @@
+describe "Import local custodian codes task" do
+  describe "import:local_custodian_codes" do
+    it "should update local_custodian_code field for LocalAuthority with correct code" do
+      args = Rake::TaskArguments.new(%i[filename], ["spec/lib/local-links-manager/import/fixtures/addressbase-local-custodian-codes-test.csv"])
+      local_authority = create(:local_authority, slug: "leicestershire", local_custodian_code: nil)
+      expect { Rake::Task["import:local_custodian_codes"].execute(args) }
+        .to change { local_authority.reload.local_custodian_code }.from(nil).to("2460")
```
ChrisBAshton, line 7: This is kind of a repeat of the unit tests in local_custodian_codes_importer_spec.rb. We've already tested the functionality extensively in there, so this test can simply check that a new instance of `LocalLinksManager::Import::LocalCustodianCodesImporter` was created.

It might be tricky to do that without using `allow_any_instance_of` - [some ideas](https://github.com/alphagov/locations-api/pull/14#discussion_r732869828) here.

lib/tasks/import/local_custodian_codes_from_csv.rake
```diff
@@ -0,0 +1,6 @@
+namespace :import do
```
ChrisBAshton, line 1: I think the rake task and associated tests can be added in a commit after the `LocalCustodianCodesImporter` has been created. Nice self-contained commits that layer functionality on - makes it easier to review and track changes.

addressbase-local-custodian-codes-street.csv
```diff
@@ -0,0 +1,393 @@
+﻿Local Custodian Code,Authority,Region,Country
```
ChrisBAshton: Similarly, I think the CSVs can be added in their own self-contained commit. The commit would ideally describe how you created these CSVs too, or where you downloaded them from.

I also think they should be grouped somewhere, e.g. in a `data/` folder instead of at the root of the repository.

ChrisBAshton (commented):
Nice one, @baisa - this is looking great! Just a few comments below.

spec/lib/tasks/local_custodian_codes_from_csv_spec.rb
baisa: Thank you for looking at it @ChrisBAshton.  In this case, I think it's better to check the actual change than mocks. The test looks the same, but it has a different role. This is more a feature/acceptance spec, as it checks one of the application entrypoints, while test for LocalCustodianCodeImporter is a simple unit spec. I would avoid to use mocks when they are not necessary, leaving them mostly for unit specs, where they help test classes in isolation, like in the PR you linked.

addressbase-local-custodian-codes-street.csv
baisa: Good point @ChrisBAshton, I have moved them to data folder in the separate commit.

lib/tasks/import/local_custodian_codes_from_csv.rake
baisa: Thank you for noticing this one, I will keep it in mind for the future, but perhaps will leave it in one commit this time as our docs advise to avoid rewriting commits unless really necessary?

ChrisBAshton: Are you referring to [these docs](https://gds-way.cloudapps.digital/standards/source-code.html#branching-merging-conventions)? I can see where it says that, but it's referring to commits that have already been pushed to `main`. Commits on unmerged branches, like this one, should "make liberal use of Git’s history rewriting features" 👍 

I agree that doc is a bit ambiguous. In this case though, we know that we "are very sure that no-one else will be affected by you rewriting the branch history", as there are no branches dependent on this one. I might raise a follow-up PR to the GDS Way to try and clarify this a bit (or you're welcome to!)

spec/lib/tasks/local_custodian_codes_from_csv_spec.rb
ChrisBAshton: Fair enough - thanks for looking 👍

addressbase-local-custodian-codes-street.csv
ChrisBAshton: Thanks @baisa - much easier to review now!

It's worth adding a [link to the Trello card](https://trello.com/c/OQKm9z4V/2758-add-local-custodian-codes-to-local-links-manager-5) in the commit message itself, rather than just refer to "the card", as it saves a developer having to find the PR associated with the commit, to find the Trello card.

ChrisBAshton (approved):
Thanks @baisa , the PR is looking great! I've left a couple of comments on the conversations above, which you may want to look at before merging, but they're non-blocking comments, so you can merge if you want.

One last (incredibly nitpicky!) comment: I've noticed the commit message summaries all end in a full stop `.`.  We don't usually do this. Not a big deal at all, but probably worth bearing in mind for future work, to keep the style of our commit messages consistent as an organisation.

addressbase-local-custodian-codes-street.csv
baisa: @ChrisBAshton makes sense! I have added the link.

lib/tasks/import/local_custodian_codes_from_csv.rake
baisa: I have split the commit @ChrisBAshton . Happy to raise follow-up PR 👍

ChrisBAshton (approved):
Perfect - thanks!

PR #877, opened 2022-02-21
Add migration to move service to higher local authority level

brucebolt:
Street Party Permissions in Leicestershire are handled at the county council (upper tier) level.

Currently there is a link for each lower tier authority.

Leicestershire County Council would like to display a link to their website instead. Removing the links to each lower tier authority will enable this.

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4882121)


KludgeKML (approved):
Looks okay (as long as we don't think we'll need to roll back).

PR #895, opened 2022-03-30
Add remove a service document

gclssvglx:
We should document the process of removing a service.

[Trello](https://trello.com/c/x43TzxRn/774-placeholder-retire-test-and-trace-lookup-tech-work-on-7th-april)

### DIFF
--- a/docs/remove-a-service.md
+++ b/docs/remove-a-service.md
@@ -1,0 +1,81 @@
+# Remove a service
+
+From time to time, either a service (LGSL code) is no longer valid, required or has been removed from the [Electronic Service Delivery (ESD)](https://standards.esd.org.uk) system. When this happens, we need to remove the service from `local-links-manager` and `publisher`.
+
+## Before you start
+
+Before starting the process it is worth gethering a few key pieces of information:
+
+* The Local Government Services List (LGSL) code of the service you are removing
+* The URL or slug of the service
+* The ESD URL of the service
+
+## 1. Back-up the links
+
+You may want to back-up the links created for the service before deleting the service.
+
+First run the [`export:links:all`](../lib/tasks/export/link_exporter.rake) rake task (via the console or [`jenkins`](#links)) which exports all of the links for all services in `local-links-manager` to the following file `public/data/links_to_services_provided_by_local_authorities.csv`. If you're running this locally, note that this file is ignored by [`.gitignore`](../.gitignore).
+
+```bash
+bundle exec rake export:links:all
+```
+
+Then retrieve the newly created file using `scp-pull`...
+
+```bash
+gds govuk connect scp-pull -e [integration|staging|production] backend source_file destination_file
+```
+
+Apps on `backend` can be found in `/var/apps/APP-NAME`. So, our data file should be available at `/var/apps/local-links-manager/public/data/links_to_services_provided_by_local_authorities.csv`.
+
+For example:
+
+```bash
+gds govuk connect scp-pull -e integration backend /var/apps/local-links-manager/public/data/links_to_services_provided_by_local_authorities.csv ~/Desktop
+```
+
+**Note**: Once stored in a secure location, please remember to delete the downloaded file from your local machine.
+
+## 2. Unpublish or update the service in `publisher`
+
+There are two options for maintaining the service going forward:
+
+1. Update the format of the service so that it becomes a standard plain text page - without a start button and local lookup functionality. Then update the page content as appropriate.
+
+2. Un-publish the service and - if desired - redirect the page.
+
+Ask a content designer to carry out the required option.
+
+## 3. Remove the LGSL code from `publisher`
+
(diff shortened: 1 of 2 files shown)


docs/remove-a-service.md
```diff
@@ -0,0 +1,79 @@
...
+Then, to retrieve the newly created file use `scp-pull`...
+
+```bash
+gds govuk connect scp-pull -e [integration|staging|production] backend source_file destination_file
+```
+
+For example:
+
+```bash
+gds govuk connect scp-pull -e integration backend /var/apps/local-links-manager/public/data/links_to_services_provided_by_local_authorities.csv ~/Desktop
```
leenagupte, line 34: Thanks for this! I love a working example 😍

docs/remove-a-service.md
```diff
@@ -0,0 +1,79 @@
...
+
+```bash
+bundle exec rake local_transactions:remove_old_services
+```
+
+## 6. Unpublish the service in `publisher`
+
+We need to remove and redirect the old `publisher` `local-transaction` service - now an `answer`.
+
+To do this, ask a content designer to `unpublish` giving them the new redirect url in `publisher`.
```
leenagupte: I thought changing the document type from local transaction to an answer doesn't affect the base path. so why do we need to do this?

docs/remove-a-service.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Check** the old URL - the new URL
+
+## 7. Remove the service from ESD
+
+Contact [Nicki Gill (Project Manager)](mailto:nicki.gill@esd.org.uk) and them to remove the service from ESD. This may take a while as it seems ESD is only updated sporadically. This is fine as there is no link between ESD and either `local-links-manager` or `publisher`.
```
leenagupte: Hmmm I'm not sure we should be putting someone's contact information on public GitHub. 😬

leenagupte (commented):
Makes sense! Thanks for adding these docs. 🥇 

I've left a couple of inline comments for clarification, but nothing major.

docs/remove-a-service.md
gclssvglx: Yes, wasn't sure about this. Will remove.

docs/remove-a-service.md
gclssvglx: To be honest, I'm not sure 🤷 I think last time we just unpublished it in `publisher` and redirected it. Could be coz we want to keep the page at the same URL - so, in this case, we won't need the un-publish step. Have asked in the slack channel in case someone knows.

gclssvglx: Yeah, seems in this instance we want to keep the existing page and URL and update the content. But for other services it might be unpublished and redirected. 

As in both cases this aspect can be done by a content designer - I'll update the wording in the doc.

docs/remove-a-service.md
```diff
@@ -0,0 +1,82 @@
...
+
+**Note**: Once stored in a secure location, please remember to delete the downloaded file from your local machine.
+
+## 2. Unpublish or update the service in `publisher`
+
+There are two options for maintaining the service going forward:
+
+1. Update the format of the service so that becomes a standard plain text page - without a start button and local lookup functionality. Then update the page content as appropriate.
+
+2. Un-publish the service and supply a redirect URL.
```
leenagupte: What are we supposed to do with the redirect URL if it's provided?

leenagupte (commented):
I have one remaining question about the redirect url. The rest makes sense and is easy to follow.

docs/remove-a-service.md
gclssvglx: I've updated the doc - and removed the redirect URL from the `Before you start` section. We (as developers) don't do anything with this URL and we don't even need to know if it exists or not to complete our tasks. Hope this clarify's things a bit 🤞

gclssvglx: Just to complicate things a little the [redirect URL is also optional](https://github.com/alphagov/publisher/blob/a20a5872088eec113b16d66585070aeda2611042/app/controllers/editions_controller.rb#L221).

docs/remove-a-service.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Note**: Once stored in a secure location, please remember to delete the downloaded file from your local machine.
+
+## 2. Unpublish or update the service in `publisher`
+
+There are two options for maintaining the service going forward:
+
+1. Update the format of the service so that it becomes a standard plain text page - without a start button and local lookup functionality. Then update the page content as appropriate.
+
+2. Un-publish the service and - if desired - supply a redirect URL.
```
leenagupte: As we don't care about the URL, perhaps it should say: "redirect the page" rather than "supply a redirect URL"?

leenagupte (approved):
One suggestion. It's not a blocker.

Thanks for writing these instructions, they'll be really helpful to a future developer ⭐

docs/remove-a-service.md
gclssvglx: Thanks for the suggestion @leenagupte - great idea 👏  Fixed up.

Rosa-Fox (commented):
Amazing! Thanks @gclssvglx!

PR #935, opened 2022-07-27
Remove Mapit dependency from rake tasks

ChrisBAshton:
Changes the `import:local_authorities:import_all` rake task to import data from `data/local-authorities.csv`, which we will manually maintain going forward. This replaces the previous logic which fetched data from Mapit.

As that removes Local Links Manager's last dependency on Mapit, all remaining references to Mapit have been removed here. As a side bonus, the new CSV import mechanism supports local custodian codes natively, so we can also remove all the local custodian code import logic we added in #850 🎉 

The rake task behaviour should be unchanged, e.g. it will error if it detects an 'orphaned' local authority, or if it detects any local authorities in the database that are missing from the CSV.

As I've removed some no-longer-existing local authorities to appease the CSV validator (rather than satisfy it with nonsense data), I was expecting the rake task to show some sort of error message for the removed authorities - but it seemed quite happy: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/304813/console

NB, it doesn't delete the missing local authorities - it skips them. This seems to be intended behaviour, defined 6 years ago in b27fb473ac0d63c8042864d1d9f76470ed3aeafc:

> We want to know if we've previously imported services, interactions or local
> authorities which subsequently disappear from the CSV.  We can't delete these
> automatically as we'll have to delete a load of dependent `ServiceInteraction`s
> and potentially useful links.  This would also have a knock on for any
> published `local_transaction` pages we have published.
>
> Because we don't know at this point what this would mean, we'll just alert and
> decide what to do later. 

Trello: https://trello.com/c/ta3NoTNg/2947-replace-mapit-powered-rake-tasks-in-local-links-manager-5


KludgeKML (approved):
LGTM (have run test suite, reset DB and run the import task and confirmed 402 Local Authority records created).

catalinailie (approved):

PR #1043, opened 2023-02-07
Add Authority active end date and support for Buckinghamshire Council merge

KludgeKML:
Adds two new columns to the LocalAuthority DB: active_end_date and active_note. These support the requirement that certain councils have been and will be merged into new unitary authorities.

Local Authorities now have an `active?` method which allows easy decisions based on whether a particular local authority is now retired. It is assumed at the moment that all such authorities will have a parent authority. When a local authority is inactive, searches for the authority itself or for links in that authority will now return values from the parent authority instead. There are utility search methods `find_current_by_slug` and `find_current_by_local_custodian_code` which take care of the active check and otherwise behave the same as `find_by(slug:)` and `find_by(local_custodian_code:)`

Following on from this change we display the status in the admin UI - in the list of councils, inactive councils are marked with a strikethrough and include the inactive_note, and in the individual authority UI, all authorities now show their status and their parent (if any), and inactive ones show the date they became inactive and the note to show why.

https://trello.com/c/EFnqnTf3/1793-buckinghamshire-add-ability-to-retire-disable-councils-to-support-council-merges-in-april

## Screenshots:
(Please ignore link flags in council pages - one set of screenshots taken on Prod, one on Dev, so no link check report available)

### Before - List of Councils
index-before

### After - List of Councils
index-after

### Before - Council Pages
Wychavon-before
Wycombe-before

### After - Council Page (active)
Wychavon-after

### After - Council Page (inactive)
Wycombe-after

### DIFF
--- a/app/views/local_authorities/index.html.erb
+++ b/app/views/local_authorities/index.html.erb
@@ -26,10 +26,13 @@
     <% @authorities.each do |authority| %>
       <tr>
         <td>
-          <%= link_to authority.name, local_authority_path(authority.slug, filter: 'broken_links'), class: 'js-open-on-submit name' %>
+          <%= link_to authority.name, local_authority_path(authority.slug, filter: 'broken_links'), class: "js-open-on-submit name #{ "authority-status-inactive" if !authority.active? }" %>
           <div class="center count <%= singular_or_plural(authority.broken_link_count) %>">
             <span><%= authority.broken_link_count %></span>
           </div>
+          <% unless authority.active? %>
+            <div class="authority-active-note add-top-padding"><%= authority.active_note %></div>
+          <% end %>
         </td>
       </tr>
     <% end %>
--- a/app/views/shared/_local_authority_details.html.erb
+++ b/app/views/shared/_local_authority_details.html.erb
@@ -14,6 +14,22 @@
     <span class="text-muted"><%= authority.homepage_link_last_checked %></span>
 
     <div>
+      <h3>Status</h3>
+      <div>Current status: <%= authority.active? ? 'active' : 'inactive' %></div>
+      <% unless authority.active? %>
+        <div>Date authority became inactive: <%= authority.active_end_date %></div>
+      <% end %>
+      <% unless authority.active_note.blank? %>
+        <div>Inactive reason: <%= authority.active_note %></div>
+      <% end %>
+      <% if authority.parent_local_authority %>
+        <div>Parent authority: <%= link_to(authority.parent_local_authority.name, local_authority_path(authority.parent_local_authority)) %></div>
+      <% end %>
+      <div>
+    </div>
+
+
+    <div>
       <h3>Download links</h3>
       <p>Produces a CSV which can be altered and uploaded to update your links as a batch.</p>
       <p>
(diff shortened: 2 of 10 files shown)


app/views/local_authorities/index.html.erb
```diff
@@ -26,10 +26,13 @@
     <% @authorities.each do |authority| %>
       <tr>
         <td>
-          <%= link_to authority.name, local_authority_path(authority.slug, filter: 'broken_links'), class: 'js-open-on-submit name' %>
+          <%= link_to authority.name, local_authority_path(authority.slug, filter: 'broken_links'), class: "js-open-on-submit name authority-status-#{authority.active? ? 'active' : 'inactive'}" %>
           <div class="center count <%= singular_or_plural(authority.broken_link_count) %>">
```
jon-kirwan, line 30: Very small suggestion but since we don't need a selector for `authority-status-active` - I wonder if the above should only add the 'inactive' CSS?

`"js-open-on-submit name #{"authority-status-inactive" if !authority.active?}"`

app/views/local_authorities/index.html.erb
```diff
@@ -26,10 +26,13 @@
...
             <span><%= authority.broken_link_count %></span>
           </div>
+          <% unless authority.active? %>
+            <div class="authority-active-note"><%= authority.active_note %></div>
+          <% end %>
```
jon-kirwan, line 35: Could be worth adding some additional padding between the heading and the change note. `govuk_admin_template` includes a few helpers that we could use:

`<div class="authority-active-note add-top-padding"><%= authority.active_note %></div>`

app/views/shared/_local_authority_details.html.erb
```diff
@@ -13,6 +13,20 @@
     <span class="<%= authority.label_status_class %> text-muted"><b><%= authority.homepage_status %></b></span>
     <span class="text-muted"><%= authority.homepage_link_last_checked %></span>
 
+    <div>
+      <h3>Status</h3>
+      <div>Current status: <%= authority.active? ? 'active' : 'inactive' %></div>
+      <% unless authority.active? %>
+        <div>Date authority became inactive: <%= authority.active_end_date %></div>
+        <div>Inactive reason: <%= authority.active_note %></div>
```
jon-kirwan, line 23: The inactive label is displayed whenever an authority is inactive. Should we also only display the label and note if an `active_note` has been added?

jon-kirwan (commented):
Reviewed the front end and all looks good to me and works as expected (I've just added a few small suggestions)

jon-kirwan:
I've noticed that the layout can look a bit crowded, particularly on smaller screens, if too much text is added in to the change note but I think this is beyond the scope of the current work and would need design input to make improvements and could be a future change.

KludgeKML:
@jon-kirwan I've implemented your changes, thanks for those! So apart from the last element I think we're good - and since we currently control what goes into those notes (they're just brought in via a rake task), I think it's safe to put that off for the moment.

jon-kirwan:
> @jon-kirwan I've implemented your changes, thanks for those! So apart from the last element I think we're good - and since we currently control what goes into those notes (they're just brought in via a rake task), I think it's safe to put that off for the moment.

Thanks for the changes @KludgeKML - all looks good front end-wise.

jon-kirwan (approved):

PR #1050, opened 2023-03-07
Add succeeded_by field to local authorities

kashifatcha:
This change is needed to handle the situation where future merges are recorded but they're into a council that isn't the current parent (for instance Cumbria, which is being split into two Unitary Authorities in April 2023 and the District Councils shared between them. They'll now show Cumbria up to the switch point, and the relevant Unitary Authorities after that.

Doing this at this point so that the Somerset merge can follow this pattern and be a simpler PR reference for the following merges.

https://trello.com/c/NyIDiB39/1844-merge-somerset-west-taunton-council

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -11,6 +11,7 @@ class LocalAuthority < ApplicationRecord
 
   has_many :links, dependent: :destroy
   belongs_to :parent_local_authority, class_name: "LocalAuthority", inverse_of: false
+  belongs_to :succeeded_by_local_authority, class_name: "LocalAuthority", inverse_of: false
   has_many :service_tiers, foreign_key: :tier_id, primary_key: :tier_id, inverse_of: :local_authority, dependent: :restrict_with_error
   has_many :services, through: :service_tiers
 
@@ -56,7 +57,7 @@ class LocalAuthority < ApplicationRecord
     return nil unless la
     return la if la.active?
 
-    la.parent_local_authority
+    la.succeeded_by_local_authority
   end
 
   def self.find_current_by_local_custodian_code(local_custodian_code)
@@ -64,6 +65,6 @@ class LocalAuthority < ApplicationRecord
     return nil unless la
     return la if la.active?
 
-    la.parent_local_authority
+    la.succeeded_by_local_authority
   end
 end
--- a/app/presenters/local_authority_presenter.rb
+++ b/app/presenters/local_authority_presenter.rb
@@ -8,4 +8,22 @@ class LocalAuthorityPresenter < SimpleDelegator
   def homepage_link_last_checked
     homepage_url.blank? ? "" : last_checked
   end
+
+  def authority_status
+    return "inactive" unless active?
+    return "active, but being retired" if active_end_date
+
+    "active"
+  end
+
+  def should_display_end_notes?
+    return true unless active?
+    return true if active? && active_end_date
+
+    false
+  end
+
+  def active_end_date_title
+    active? ? "Date authority is due to become inactive" : "Date authority became inactive"
+  end
 end
--- a/spec/models/local_authority_spec.rb
+++ b/spec/models/local_authority_spec.rb
@@ -95,10 +95,10 @@ RSpec.describe LocalAuthority, type: :model do
       expect(LocalAuthority.find_current_by_slug(local_authority.slug)).to eq local_authority
     end
 
-    it "returns the parent authority if the authority is inactive" do
-      parent_authority = create(:county_council)
-      local_authority = create(:district_council, parent_local_authority: parent_authority, active_end_date: Time.zone.now - 1.year)
-      expect(LocalAuthority.find_current_by_slug(local_authority.slug)).to eq parent_authority
+    it "returns the succeeded_by authority if the authority is inactive" do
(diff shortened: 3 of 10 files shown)


app/presenters/local_authority_presenter.rb
```diff
@@ -8,4 +8,19 @@ def homepage_status
...
+
+  def display_status
+    return "inactive" unless active?
+    return "active, but being retired" if active_end_date
+
+    "active"
+  end
+
+  def display_end_notes?
+    !(active? && active_end_date.nil?)
```
1pretz1: This conditional could be simplified slightly (`nil` is false in Ruby world):

```
!active? && active_end_date
```

spec/models/local_authority_spec.rb
```diff
@@ -96,9 +96,9 @@
     end
 
     it "returns the parent authority if the authority is inactive" do
```
1pretz1: Should this be renamed to "succeeded_by" authority?

app/presenters/local_authority_presenter.rb
```diff
@@ -8,4 +8,19 @@ def homepage_status
...
+  end
+
+  def display_end_title
```
1pretz1: Feels like this should just be `end_title` and same comment for `display_status` -> `status`

app/presenters/local_authority_presenter.rb
```diff
@@ -8,4 +8,19 @@ def homepage_status
   def homepage_link_last_checked
     homepage_url.blank? ? "" : last_checked
   end
+
```
1pretz1, line 11: We should probably add unit tests for the added functionality as we have a spec for this presenter: https://github.com/alphagov/local-links-manager/blob/main/spec/presenters/local_authority_presenter_spec.rb.

We also could add / update the feature test: https://github.com/alphagov/local-links-manager/blob/main/spec/features/local_authorities/local_authority_show_spec.rb.

app/models/local_authority.rb
```diff
@@ -56,14 +57,14 @@ def self.find_current_by_slug(slug)
     return nil unless la
     return la if la.active?
 
-    la.parent_local_authority
+    la.succeeded_by_local_authority
```
1pretz1, line 60: Should this be a conditional, e.g if `succeeded_by_local_authority` exists then return, `else` `parent_local_authority` as I noticed we're not populating the new field, bar for `buckinghamshire` merge?  (same comment for below method).

1pretz1 (commented):
Good job 👍 Left a few comments, one of them is probably down to lack of context / understanding.

spec/models/local_authority_spec.rb
KludgeKML: :+1: (there's also a set of other copieos). Fixing.

app/models/local_authority.rb
KludgeKML: No, we shouldn't be returning parent authority here - only succeeded_by, and if that's nil, we should return it as nil. It was only appropriate to return parent_authority when we were using it as a proxy for succeeding authority.

app/presenters/local_authority_presenter.rb
KludgeKML: That subtly changes the return condition. What I'm describing here is that display_end_notes? should return true if:
- the LocalAuthority is inactive (ie active? FALSE), OR
- the LocalAuthority is active, but has an active end data (ie active? TRUE, active_end_date not nil)

1pretz1: Ah okay, thanks for explaining! In that case, could the conditional be more explicit? Below costs an extra line but improves readability - what do you think? 

E.g 

```
return true if !active?
return true if active && active_end_date
```

app/models/local_authority.rb
1pretz1: Ah cool, thanks for clarifying 👍

app/presenters/local_authority_presenter.rb
KludgeKML: Cool - fixed (with mild change that it's active? in line 2 and rubocop wants an unless rather than an if ! in line 1)

app/presenters/local_authority_presenter.rb
KludgeKML: Unit tests added for the presenter, just looking at feature tests.

app/presenters/local_authority_presenter.rb
KludgeKML: status will break it, because of a dubious existing design choice for the LocalAuthority model that means we already have a status field we can't rename. But I've changed these a bit to be more in keeping with the names of the fields and hopefully more useful:

- display_status -> authority_status
- display_end_notes? -> should_display_end_notes?
- display_end_title -> active_end_date_title

app/presenters/local_authority_presenter.rb
KludgeKML: All added.

1pretz1: Awesome, thanks for that

app/presenters/local_authority_presenter.rb
1pretz1: Looks good

1pretz1 (approved):
Awesome! LGTM 🟢

PR #1080, opened 2023-05-04
Expose GSS Code in API

KludgeKML:
New local authorities are getting GSS codes, but SNAC codes are not being assigned any more. This change exposes the GSS codes in API calls so that dependent apps (eg Imminence) can use the GSS for their local authority search as well as the old SNACs. Ultimately we should look to decommission SNACS and remove them. There are a couple of additional QoL changes to help with the lack of SNACs:

- Allow SNAC to be nil (currently presence/uniqueness is validated, alter to validate uniqueness with multiple nils)
- Display SNAC/GSS in the Local Authority page so that users can look up SNAC/GSS codes more easily.

https://trello.com/c/i66BCwUl/1978-imminence-cant-perform-local-authority-searches-for-new-local-authorities

### Before: 
Screenshot 2023-05-10 at 16 37 53

### After:
Screenshot 2023-05-10 at 16 37 35


muzakh (approved):

spec/presenters/local_authority_hash_presenter_spec.rb
```diff
@@ -0,0 +1,50 @@
+describe LocalAuthorityApiResponsePresenter do
```
kashifatcha, line 1: Would it be possible to squash this commit into the previous one as this is the spec for the changes made in the previous commit. It's easier to view the changes that way.

app/views/local_authorities/show.html.erb
```diff
@@ -13,7 +13,7 @@
```
kashifatcha, line 13: As this change is frontend it would be useful to have an accompanying screenshot. Would it be possible to add one? I ran the application and the table looks good, and is working as expected. Nice work in thinking of this, as it's really useful!

spec/features/local_authorities/local_authority_show_spec.rb
```diff
@@ -13,6 +13,12 @@
     end
   end
 
+  it "display the gss code in a table" do
+    within("table#codes") do
+      expect(page).to have_content local_authority.gss
```
kashifatcha, line 18: Could you also add expectations for local_custodian_code and snacs?

spec/models/local_authority_spec.rb
```diff
@@ -6,12 +6,23 @@
...
-    it { should validate_uniqueness_of(:snac) }
     it { should validate_uniqueness_of(:slug) }
+    it { should validate_uniqueness_of(:snac) }
+
+    describe "snac" do
+      context "with multiple records with nil snac" do
+        before do
+          create(:local_authority, snac: nil)
+        end
+        it "saves all records, allowing multiple nils" do
```
kashifatcha: To avoid repetition it looks like the spec from line 21 doesn't need to be here as it is already covered in line 10.

kashifatcha (commented):
Looks really good! Nice work. Just left a few minor in-line comments.

spec/presenters/local_authority_hash_presenter_spec.rb
KludgeKML: Done!

app/views/local_authorities/show.html.erb
KludgeKML: Done

spec/models/local_authority_spec.rb
KludgeKML: Okay, removed (allow nil skips other validations, so no need to confirm nil isn't affected by uniqueness), thanks to @1pretz1 

> https://edgeguides.rubyonrails.org/active_record_validations.html#allow-nil

kashifatcha (approved):
LGTM

PR #1083, opened 2023-05-10
Fix cookie overflow on multiple errors

KludgeKML:
If there are more than 5 errors in import, show only the first 5.

- All errors are identical, so we can save space in the cookie by showing the error once, then a list of line numbers. 50 should fit into the cookie easily.
- Prevents cookie overflow if many lines in the import are broken.
- If all lines are broken, adds a message to that effect.
- If no errors are detected but nothing is imported, warns that there might be a formatting issue that isn't an error but is preventing import.

https://trello.com/c/fZhH3A5O/1995-ps-11-local-links-manager-crashes-with-cookie-overflow-when-there-are-too-many-import-errors


spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -63,18 +63,73 @@
...
+      end
+    end
 
-      expect(response.status).to eq(302)
-      expect(response.location).to match(url_regex)
-      expect(response.headers["Content-Type"]).to eq("text/html; charset=utf-8")
+    context "with errors in the CSV" do
+      before do
+        @local_authority = create(:local_authority)
+        service1 = create(:service, lgsl_code: 1)
```
1pretz1: Could do a loop here which DRYs it up a bit, something like the below:
```
interaction = create(:interaction, lgil_code: 1)

6.times do |i|
  service = create(:service, lgsl_code: i)
  create(:service_interaction, service: service1, interaction:)  
end
```

spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -63,18 +63,73 @@
...
+        expect(response.headers["Content-Type"]).to eq("text/html; charset=utf-8")
+      end
+    end
 
-      expect(response.status).to eq(302)
-      expect(response.location).to match(url_regex)
-      expect(response.headers["Content-Type"]).to eq("text/html; charset=utf-8")
+    context "with errors in the CSV" do
+      before do
+        @local_authority = create(:local_authority)
```
1pretz1: It is a bit strange we're using instance variables vs let in RSpec files, I can see you're following the pattern in the file but maybe you can change all the instance vars to `let`?

spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -63,18 +63,73 @@
...
+        interaction = create(:interaction, lgil_code: 1)
+        create(:service_interaction, service: service1, interaction:)
+        create(:service_interaction, service: service2, interaction:)
+        create(:service_interaction, service: service3, interaction:)
+        create(:service_interaction, service: service4, interaction:)
+        create(:service_interaction, service: service5, interaction:)
+        create(:service_interaction, service: service6, interaction:)
+      end
+
+      let(:path_all) { Rails.root.join("spec/lib/local-links-manager/import/fixtures/imported_links_all_errors.csv") }
```
1pretz1: I think we only need 4 lets (csv_*) as the `path_` are only used another let rather than in any of the specs, e.g:
```
let(:csv_all) do
  path_all = Rails.root.join("spec/lib/local-links-manager/import/fixtures/imported_links_all_errors.csv")
  Rack::Test::UploadedFile.new(path_all, "text/csv", true)
end 
```

Although arguably, we probably should just include the individual files in the `it` blocks themselves as they aren't shared between tests.

spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -63,18 +63,73 @@
...
+      let(:csv_all) { Rack::Test::UploadedFile.new(path_all, "text/csv", true) }
+      let(:path_many) { Rails.root.join("spec/lib/local-links-manager/import/fixtures/imported_links_many_errors.csv") }
+      let(:csv_many) { Rack::Test::UploadedFile.new(path_many, "text/csv", true) }
+      let(:path_few) { Rails.root.join("spec/lib/local-links-manager/import/fixtures/imported_links_few_errors.csv") }
+      let(:csv_few) { Rack::Test::UploadedFile.new(path_few, "text/csv", true) }
+      let(:path_nothing) { Rails.root.join("spec/lib/local-links-manager/import/fixtures/imported_links_nothing_to_import.csv") }
+      let(:csv_nothing) { Rack::Test::UploadedFile.new(path_nothing, "text/csv", true) }
+      let(:url_regex) { /http:\/\/.+\/local_authorities\/#{@local_authority.slug}/ }
```
1pretz1: Don't think we're using `url_regex` in this context

spec/controllers/local_authorities_controller_spec.rb
```diff
@@ -63,18 +63,73 @@
...
+
+      it "shows the all error message if all lines are broken" do
+        login_as_stub_user
```
1pretz1: We could add `login_as_stub_user` to the `before` block rather than include in each test

1pretz1 (commented):
Looks good and much improved error reporting, nice! Left a few comments regarding the spec but nothing major

spec/controllers/local_authorities_controller_spec.rb
KludgeKML: Done

spec/controllers/local_authorities_controller_spec.rb
KludgeKML: I've moved them into the tests, with a let for the base path.

spec/controllers/local_authorities_controller_spec.rb
KludgeKML: Agreed, made into a let.

spec/controllers/local_authorities_controller_spec.rb
KludgeKML: Nice, done (with the caveat that it's lgsl_code: i+1)

1pretz1 (approved):

PR #1222, opened 2024-01-31
Remove roll-your-own cache methods

KludgeKML:
There's a roll-your-own cache key method, and it's used to cache view fragments that don't really need that much caching. Having deployed this to integration there's no particular slowdown in the UI, and the UI is rarely used anyway.

https://trello.com/c/F7mchyKG/1605-ps-15-investigate-caching-in-local-links-manager-admin-console


1pretz1 (approved):
Nice, thanks for jumping on and testing! Glad the outcome was that we don't need the bespoke caching at all

PR #1249, opened 2024-02-28
Homepage problems report

KludgeKML:
- Add a column to the local authorities index to show homepage status.
- Add filters to the local authorities index to allow toggling only active councils (on by default) and only councils with homepage problems.

https://trello.com/c/zoxV3oIw/2376-report-on-la-homepage-warnings

We can use this filter pattern on the other listings pages.

Search filter is still the one used by specialist finders and signon (which is nice and snappy). It does mean that there's a slight behaviour difference (in page filter works as you type, the two checkbox filters only work when you click Update), but this is probably fine for a govuk-only facing application?

## BEFORE

Screenshot 2024-02-28 at 13 12 53

## AFTER

Screenshot 2024-02-29 at 15 24 21

### DIFF
--- a/app/views/local_authorities/_filter.html.erb
+++ b/app/views/local_authorities/_filter.html.erb
@@ -1,0 +1,32 @@
+
+
+<%= form_with url: local_authorities_path, method: :get do |form| %>
+  <%= render "govuk_publishing_components/components/option_select", {
+    options_container_id: "list_of_vegetables",
+    title: "Only show",
+    key: "filter",
+    options: [
+      {
+        value: "only_active",
+        label: "Active councils",
+        checked: params[:filter]&.include?("only_active"),
+      },
+      {
+        value: "only_homepage_problems",
+        label: "Homepage problems",
+        checked: params[:filter]&.include?("only_homepage_problems"),
+      },
+    ],
+  } %>
+
+  <%= render "govuk_publishing_components/components/button", {
+    text: "Update",
+    margin_bottom: 4,
+    data_attributes: {
+      module: "gem-track-click",
+      "track-category": "form-button",
+      "track-action": "local-authority-filter-button",
+      "track-label": "Update",
+    },
+  } %>
+<% end %>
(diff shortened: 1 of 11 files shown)


app/views/local_authorities/_filter.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      items: [
+        {
+          label: "Only with homepage problems",
+          value: "1",
+          bold: true,
+          checked: params["only_homepage_problems"],
+        },
+      ],
+    } %>
+
```
jon-kirwan: I think we might have a ready pattern we can use here in the option select component https://components.publishing.service.gov.uk/component-guide/option_select.

app/views/local_authorities/_filter.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="app-view-filter govuk-!-margin-right-5 govuk-!-padding-5">
+  <%= form_with url: local_authorities_path, method: :get do |form| %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Filter by",
+      margin_bottom: 4,
+    } %>
+
```
jon-kirwan, line 1: Maybe, like with search pages https://www.gov.uk/aaib-reports, we can include a visually hidden heading here and save some more space
`<h2 class="govuk-visually-hidden">Filter</h2>`

app/views/local_authorities/_filter.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="app-view-filter govuk-!-margin-right-5 govuk-!-padding-5">
```
jon-kirwan: I think we could remove the `govuk-!-margin-right-5` class to give a little more space.

KludgeKML: Not using the app-view-filter block any more.

app/views/local_authorities/_filter.html.erb
KludgeKML: Using this now.

app/views/local_authorities/_filter.html.erb
KludgeKML: Header is smaller now, part of the options select.

jon-kirwan (approved):
Thanks Keith, looks good to me!

PR #1266, opened 2024-03-18
Publish external content

KludgeKML:
Local Links Manager should publish council homepages as External Content items to publishing API, to take over the work currently done by SearchAdmin (which currently has older data, and generally has worse support for automated link checking and maintenance).

- Add a content_id field to Local Authority records - this is the content_id of the external content item related to their homepage.
- Add callbacks so that when a Local Authority is updated, the external content item will be created / updated / unpublished (if council is retired).
- Add a rake task to do initial publishing of all items (all active council's homepages).

There's an equivalent search-admin task to remove all the council external links from Search Admin which should be run before the task in Local Links Manager.

https://trello.com/c/p6bYucUh/2404-local-links-manager-should-publish-council-homepages-to-the-search-api

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -1,4 +1,6 @@
 class LocalAuthority < ApplicationRecord
+  after_commit :update_external_content, if: :persisted?
+
   validates :gss, :slug, uniqueness: true
   validates :snac, uniqueness: true, allow_nil: true
   validates :gss, :name, :slug, presence: true
@@ -69,5 +71,15 @@ class LocalAuthority < ApplicationRecord
     return la if la.active?
 
     la.succeeded_by_local_authority
+  end
+
+private
+
+  def update_external_content
+    if active?
+      LocalAuthorityExternalContentPublisher.publish(self)
+    else
+      LocalAuthorityExternalContentPublisher.unpublish(self)
+    end
   end
 end
--- a/app/presenters/local_authority_external_content_presenter.rb
+++ b/app/presenters/local_authority_external_content_presenter.rb
@@ -1,0 +1,19 @@
+class LocalAuthorityExternalContentPresenter
+  def initialize(local_authority)
+    @local_authority = local_authority
+  end
+
+  def present_for_publishing_api
+    {
+      description: "Website of #{@local_authority.name}",
+      details: {
+        url: @local_authority.homepage_url,
+      },
+      document_type: "external_content",
+      publishing_app: "local-links-manager",
+      schema_name: "external_content",
+      title: @local_authority.name,
+      update_type: "minor",
+    }
+  end
+end
--- a/db/migrate/20240318105241_add_content_id_to_local_authorities.rb
+++ b/db/migrate/20240318105241_add_content_id_to_local_authorities.rb
@@ -1,0 +1,8 @@
+class AddContentIdToLocalAuthorities < ActiveRecord::Migration[7.1]
+  def change
+    enable_extension "pgcrypto"
+
+    add_column :local_authorities, :content_id, :uuid, default: "gen_random_uuid()"
+    add_index :local_authorities, :content_id, unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,8 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_05_04_115751) do
+ActiveRecord::Schema[7.1].define(version: 2024_03_18_105241) do
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pgcrypto"
   enable_extension "plpgsql"
 
   create_table "interactions", id: :serial, force: :cascade do |t|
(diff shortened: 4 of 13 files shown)


db/migrate/20240318105241_add_content_id_to_local_authorities.rb
```diff
@@ -0,0 +1,14 @@
+class AddContentIdToLocalAuthorities < ActiveRecord::Migration[7.1]
+  def up
+    add_column :local_authorities, :content_id, "char(36)"
```
csutter: Given this is Postgres and I'm assuming we're unlikely to ever migrate to anything else, using the native UUID type might make more sense. We could even have Postgres generate it for us if we wanted to:

```ruby
enable_extension "pgcrypto"

add_column :local_authorities, :content_id, :uuid, default: "gen_random_uuid()"
```

(code off the top of my head so do double check it works!)

That way we don't need to think about generating/assigning it app-side.

app/presenters/local_authority_external_content_presenter.rb
```diff
@@ -0,0 +1,20 @@
+class LocalAuthorityExternalContentPresenter
+  def initialize(local_authority)
+    @local_authority = local_authority
+  end
+
+  def present_for_publishing_api
+    {
+      description: "Website of #{@local_authority.name}",
+      details: {
+        hidden_search_terms: [@local_authority.name],
```
csutter: I want to say we can probably get rid of `hidden_search_terms` for the purposes of Search v2, given that it's exactly the same as the title anyway 🤔

KludgeKML: Done

db/migrate/20240318105241_add_content_id_to_local_authorities.rb
KludgeKML: Yeah, I wasn't sure if we wanted to get yet another type of UUID into publishing API, but those PG ones are just UUID v4, which is what SecureRandom is, and it allows us to get rid of some model-level validation since it'll be enforced by the database anyway.

Done

richardTowers (approved):
Looks good to me

db/schema.rb
```diff
@@ -103,6 +106,7 @@
     t.string "slug", null: false
     t.boolean "enabled", default: false, null: false
     t.integer "broken_link_count", default: 0
+    t.boolean "unofficial", default: false
```
leenagupte: What is the "unofficial" field?

app/models/local_authority.rb
```diff
@@ -1,4 +1,6 @@
 class LocalAuthority < ApplicationRecord
+  after_commit :update_external_content, if: :persisted?
```
leenagupte, line 2: Nice!

app/presenters/local_authority_external_content_presenter.rb
```diff
@@ -0,0 +1,19 @@
+class LocalAuthorityExternalContentPresenter
```
leenagupte, line 1: I feels like a test is missing for this class and ` LocalAuthorityExternalContentPublisher`

spec/models/local_authority_spec.rb
```diff
@@ -144,4 +144,54 @@
...
+        stub_publishing_api_for_subject(subject)
+        subject.save!
+        WebMock.reset!
+
+        subject.homepage_url = "https://www.example.com/active-council"
+        stubs = stub_publishing_api_for_subject(subject)
+        subject.save!
+
+        expect(stubs.first).to have_been_requested.once
+        expect(stubs.last).to have_been_requested.once
```
leenagupte, line 162: Are these testing the same thing? i.e. just line 158 and not that line 153 is also being executed?

leenagupte: I love the tests though 🎉

leenagupte (commented):
Overall this looks really good. 🎉 

I have a couple of inline comments about the tests.

db/schema.rb
leenagupte: I think this might be a hangover from a local commit in a different [draft PR](https://github.com/alphagov/local-links-manager/pull/1247/commits/5f32a20169a0382d1e5b25ea86a134d51738b834)

KludgeKML: Yeah - I thought I'd caught this slippery blighter when I was rebasing, but apparently it got past me again! Thanks!

spec/models/local_authority_spec.rb
KludgeKML: stubs (returned from the stub_publishing_api_for_subject) is an array (because the underlying test helper from gds-api-adapters stubs all three of the apis potentially required to publish an item: 0 = put the item, 1 = add links, 2 = publish. So we want to test that it both created and published the item (0 and 2, or first and last). We don't create links. I could test that that one isn't called, but that feels like overkill.

app/presenters/local_authority_external_content_presenter.rb
KludgeKML: Hmm. So, both of these classes are 100% covered by the model callback tests, and I feel like that's okay for the publisher class because all that's doing is abstracting the api calls, so probably unit test doesn't add anything. 

However, thinking this through to defend it, I think the Presenter class probably does want a unit test, because it's used as both test value creation and code path in the feature test, so if it breaks it won't necessarily fail that test. I'll add a unit test for it.

db/schema.rb
KludgeKML: Fixed

app/presenters/local_authority_external_content_presenter.rb
KludgeKML: Done!

spec/presenters/local_authority_external_content_presenter_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+        },
+        document_type: "external_content",
+        publishing_app: "local-links-manager",
+        schema_name: "external_content",
+        title: "Angus County Council",
+        update_type: "minor",
+      }
+    end
+
+    it "returns a hash appropriate for an external content item in the Publishing API" do
```
leenagupte, line 19: 🎉

leenagupte (approved):
LGTM 👍 

Thanks for adding the test!

PR #1275, opened 2024-03-26
Fix publishing api errors

KludgeKML:
The publish/unpublish code for Local Authorities is getting failures because it tries to unpublish missing documents. This causes errors when the link checker updates the records nightly, so add code to check before calling.

Also, change the behaviour to only publish if the homepage_url or name are changed. It currently publishes when unrelated changes are made to a local authority model, which creates a bunch of spam editions given that the link checker updates each authority once a day. 

### DIFF
--- a/app/models/local_authority.rb
+++ b/app/models/local_authority.rb
@@ -76,10 +76,10 @@ class LocalAuthority < ApplicationRecord
 private
 
   def update_external_content
-    if active?
-      LocalAuthorityExternalContentPublisher.publish(self)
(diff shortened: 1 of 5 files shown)


jkempster34 (commented):
Hiya, considering this is my last day at GDS, and there is still some discussion around this on Slack, I'm not going to be able to review it. Hope that's okay

aldavidson (approved):
This looks fine to me, ship it and we'll see if it stops the 409s

PR #1291, opened 2024-04-28
Improve code coverage to 95%

georges1996:


lib/tasks/once_off/publish_external_content_items.rake
```diff
@@ -1,8 +0,0 @@
-namespace :once_off do
```
georges1996, line 1: As most of the code coverage is due to some rake tasks not being fully tested I have removed the `:once_off ` ones as it seems these are only needed to be ran once and thats it? If we can remove these it aids with improving the code coverage, but let me know if we still need these.

lib/tasks/once_off/remove_unused_data.rake
```diff
@@ -1,30 +0,0 @@
-namespace :once_off do
```
georges1996, line 1: As most of the code coverage is due to some rake tasks not being fully tested I have removed the :once_off  ones as it seems these are only needed to be ran once and thats it? If we can remove these it aids with improving the code coverage, but let me know if we still need these.

lib/tasks/once_off/wrong_tier_links.rake
```diff
@@ -1,10 +0,0 @@
-namespace :once_off do
```
georges1996, line 1: As most of the code coverage is due to some rake tasks not being fully tested I have removed the :once_off  ones as it seems these are only needed to be ran once and thats it? If we can remove these it aids with improving the code coverage, but let me know if we still need these.

PR #1362, opened 2024-07-30
Allow Per-department permissions

KludgeKML:
Adds the ability to mark a service as owned by one or more organisations, allowing users from those organisations that have been granted access to Local Links Manager in signon to view and edit links for those services. Other existing actions, and the ability to set owners, now only available with the new `GDS Editor` permission.

A menu option to Switch app has been added, consistent with Whitehall.

New tests are Request/System specs - we have not updated old Controller/Feature tests in this app, that has been captured as technical debt for updating later.

⚠️ **WARNING: merging PR without first updating existing GDS users of LLM with the new permission will result in them losing the ability to view local authorities and services. Update their permissions before merging!** ⚠️

- [x] Update permissions

https://trello.com/c/1eUXHHBz/22-restrict-access-to-a-specific-service, [Jira issue PNP-8560](https://gov-uk.atlassian.net/browse/PNP-8560)

### View of the services page when logged in as a user (note main menu does not offer the broken links or council options)

Screenshot 2024-08-14 at 11 24 26

### View of an owned service page when logged in as a user (note main menu does not offer the broken links or council options, update owners link is not available in sidebar)

Screenshot 2024-08-14 at 11 24 34

### View of the edit link page for an owned service page when logged in as a user (note main menu does not offer the broken links or council options, delete option is not available)

Screenshot 2024-08-14 at 11 24 43

### View of an owned service page when logged in with GDS Editor permissions (update owners link is available in sidebar)

Screenshot 2024-08-14 at 11 22 05

### New Update Owners form, available on when logged in with GDS Editor permissions 

Screenshot 2024-08-14 at 11 22 19

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,6 +1,11 @@
 class LinksController < ApplicationController
   before_action :load_dependencies, only: %i[edit update destroy]
   before_action :set_back_url_before_post_request, only: %i[edit update destroy]
+
+  before_action :forbid_unless_permission, only: %i[edit update]
+  before_action :forbid_unless_gds_editor, only: %i[destroy homepage_links_status_csv links_status_csv bad_links_url_and_status_csv]
+  before_action :redirect_unless_gds_editor, only: %i[index]
+
   helper_method :back_url
 
   def index
--- a/app/controllers/local_authorities_controller.rb
+++ b/app/controllers/local_authorities_controller.rb
@@ -4,6 +4,9 @@ class LocalAuthoritiesController < ApplicationController
 
   before_action :set_authority, except: %i[index bad_homepage_url_and_status_csv]
 
+  before_action :forbid_unless_gds_editor, only: %i[update download_links_csv upload_links_csv bad_homepage_url_and_status_csv]
+  before_action :redirect_unless_gds_editor, only: %i[index show edit_url download_links_form upload_links_form]
+
   def index
     Rails.logger.info(params)
 
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -6,6 +6,8 @@ class ServicesController < ApplicationController
 
   before_action :forbid_unless_permission, except: %i[index]
 
+  helper_method :org_name_for_current_user
+
   def index
     @services = services_for_user(current_user).enabled.order(broken_link_count: :desc)
 
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -4,7 +4,8 @@ class ServicesController < ApplicationController
 
   before_action :set_service, except: :index
 
-  before_action :forbid_unless_permission, except: %i[index]
+  before_action :forbid_unless_permission, only: %i[show download_links_form download_links_csv upload_links_form upload_links_csv]
+  before_action :forbid_unless_gds_editor, only: %i[update_owner_form update_owner]
 
   helper_method :org_name_for_current_user
 
@@ -19,6 +20,16 @@ class ServicesController < ApplicationController
     @link_filter = params[:filter]
     @links = links_for_service
     @breadcrumbs = service_breadcrumbs(@service)
+  end
+
+  def update_owner_form
+    @breadcrumbs = service_breadcrumbs(@service) + [{ title: "Update Owner", url: update_owner_form_service_path(@service) }]
+  end
+
+  def update_owner
+    @service.update!(organisation_slugs: params["service"]["organisation_slugs"].split(" "))
+
+    redirect_to service_path(@service, filter: "broken_links")
   end
 
   def download_links_form
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -4,9 +4,10 @@ class ServicesController < ApplicationController
 
   before_action :set_service, except: :index
 
+  before_action :forbid_unless_permission, except: %i[index]
+
   def index
-    @services = Service.enabled.order(broken_link_count: :desc)
-    raise "Missing Data" if @services.empty?
+    @services = services_for_user(current_user).enabled.order(broken_link_count: :desc)
 
     @breadcrumbs = index_breadcrumbs
   end
@@ -34,11 +35,18 @@ class ServicesController < ApplicationController
   end
 
   def upload_links_csv
-    attempt_import(:service, @service)
-    redirect_to service_path(@service)
+    return redirect_to service_path(@service) if attempt_import(:service, @service)
+
+    redirect_to(upload_links_form_service_path(@service))
   end
 
 private
+
+  def services_for_user(user)
+    return Service.all if gds_editor?
+
+    Service.where(":organisation_slugs = ANY(organisation_slugs)", organisation_slugs: user.organisation_slug)
+  end
 
   def set_service
     @service = Service.find_by!(slug: params[:service_slug])
--- a/app/presenters/links_table_presenter.rb
+++ b/app/presenters/links_table_presenter.rb
@@ -15,7 +15,7 @@ class LinksTablePresenter
         { text: link.analytics.to_i, format: "numeric" },
         { text: "<span class=\"app-service-label\">#{title}</span>".html_safe },
         { text: "<span class=\"app-interaction-label\">#{link.interaction.label}</span>".html_safe },
-        { text: @view_context.link_to(link.local_authority.name, @view_context.local_authority_path(link.local_authority), class: "govuk-link") },
+        { text: link.local_authority.name },
         { text: "<span class=\"govuk-tag govuk-tag--#{pres.status_tag_colour}\">#{pres.status_description}</span>".html_safe },
         { text: @view_context.link_to("Edit <span class=\"govuk-visually-hidden\">#{link.local_authority.name} - #{si.service.label} - #{si.interaction.label}</span>".html_safe, @view_context.edit_link_path(link.local_authority, link.service, link.interaction), class: "govuk-link") },
       ]
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -16,11 +16,11 @@
   browser_title: (yield :page_title),
 } do %>
 
-<%= render "govuk_publishing_components/components/layout_header", {
-  product_name: "Local Links Manager",
-  environment:,
-  navigation_items: [
-    {
+<%
+  menu_option = []
+
+  if gds_editor?
+    menu_option += [{
       text: "Broken Links",
       href: root_path,
       active: current_page?(root_path),
@@ -29,13 +29,25 @@
       text: "Councils",
       href: local_authorities_path(filter: %w[only_active]),
       active: current_page?(local_authorities_path),
-    },
-    {
-      text: "Services",
-      href: services_path,
-      active: current_page?(services_path),
-    },
-  ],
+    }]
+  end
+
+  menu_option << {
+    text: "Services",
+    href: services_path,
+    active: current_page?(services_path),
+  }
+
+  menu_option << {
+    text: "Switch app",
+    href: Plek.external_url_for("signon"),
+  }
+%>
+
+<%= render "govuk_publishing_components/components/layout_header", {
+  product_name: "Local Links Manager",
+  environment:,
+  navigation_items: menu_option,
   } %>
 
 <div class="govuk-width-container">
--- a/spec/controllers/services_controller_spec.rb
+++ b/spec/controllers/services_controller_spec.rb
@@ -1,19 +1,10 @@
 RSpec.describe ServicesController, type: :controller do
   describe "GET #index" do
-    context "when there is missing data" do
-      it "returns http server error" do
-        login_as_stub_user
-        expect { get :index }.to raise_error "Missing Data"
-      end
-    end
-
-    context "when there is sufficient data" do
-      it "returns http succcess" do
-        login_as_stub_user
-        create(:service)
-        get :index
-        expect(response).to have_http_status(200)
-      end
+    it "returns http succcess" do
+      login_as_stub_user
+      create(:service)
+      get :index
+      expect(response).to have_http_status(200)
     end
   end
 
--- a/spec/controllers/services_controller_spec.rb
+++ b/spec/controllers/services_controller_spec.rb
@@ -1,7 +1,8 @@
 RSpec.describe ServicesController, type: :controller do
+  before { login_as_gds_editor }
+
   describe "GET #index" do
     it "returns http succcess" do
-      login_as_stub_user
       create(:service)
       get :index
       expect(response).to have_http_status(200)
@@ -10,7 +11,6 @@ RSpec.describe ServicesController, type: :controller do
 
   describe "GET #show" do
     it "returns http success" do
-      login_as_stub_user
       service = create(:service)
       get :show, params: { service_slug: service.slug }
       expect(response).to have_http_status(200)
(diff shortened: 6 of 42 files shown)


spec/support/authentication.rb
```diff
@@ -1,4 +1,10 @@
 module AuthenticationControllerHelpers
+  def login_as_new(user)
+    allow_any_instance_of(ApplicationController).to receive(:authenticate_user!).and_return(true)
+    allow_any_instance_of(ApplicationController).to receive(:user_signed_in?).and_return(true)
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(user)
+  end
+
   def login_as(user)
```
leenagupte: Is this method still being used? I think it was only being called by `login_as_stub_user`.

Perhaps then rather than creating a new `login_as_new` method, the existing `login_as` method could be updated to included the extra mocks?

spec/controllers/services_controller_spec.rb
```diff
@@ -1,11 +1,11 @@
...
-        login_as_stub_user
-        expect { get :index }.to raise_error "Missing Data"
-      end
-    end
+    # context "when there is missing data" do
+    #   it "returns http server error" do
+    #     login_as_stub_user
+    #     expect { get :index }.to raise_error "Missing Data"
+    #   end
+    # end
```
leenagupte: Should this still be commented out?

spec/features/services/service_show_spec.rb
```diff
@@ -43,10 +43,6 @@
   end
 
   describe "for each local authority" do
-    it "the Local Authority name is linked to the council specific page" do
```
leenagupte, line 46: Was what this is testing removed somewhere else?  I don't think this change belongs in this commit though.

app/views/layouts/application.html.erb
```diff
@@ -16,26 +16,29 @@
...
+  menu_option << {
+    text: "Broken Links",
+    href: root_path,
+    active: current_page?(root_path),
+  } if gds_editor?
+  menu_option << {
+    text: "Councils",
+    href: local_authorities_path(filter: %w[only_active]),
+    active: current_page?(local_authorities_path),
+  } if gds_editor?
```
leenagupte: Considering that these to menu options have the same conditional, I'd probably re-order the logic so that it's easier to read.

i.e.

```ruby
if gds_editor?
 menu_option << [
  {
   text: "Broken Links",
    href: root_path,
    active: current_page?(root_path),
  },
  {
    text: "Councils",
    href: local_authorities_path(filter: %w[only_active]),
    active: current_page?(local_authorities_path),
  }
 ]
end


```

spec/support/authentication.rb
```diff
@@ -11,8 +17,69 @@ def stub_user
...
+    login_as_new(create(:user, permissions: ["GDS Editor"], organisation_slug: "government-digital-service"))
+  end
+
   def login_as_stub_user
-    login_as stub_user
+    login_as_new(stub_user)
+  end
+end
+
+RSpec.shared_examples "redirects non-GDS Editors to services page" do |path|
```
leenagupte, line 17: Shared examples are great, but where are they being used?

app/presenters/links_table_presenter.rb
```diff
@@ -15,7 +15,7 @@ def rows
         { text: link.analytics.to_i, format: "numeric" },
         { text: "<span class=\"app-service-label\">#{title}</span>".html_safe },
         { text: "<span class=\"app-interaction-label\">#{link.interaction.label}</span>".html_safe },
-        { text: @view_context.link_to(link.local_authority.name, @view_context.local_authority_path(link.local_authority), class: "govuk-link") },
+        { text: link.local_authority.name },
```
leenagupte, line 18: I think this relates to the test removed in an earlier commit?

spec/requests/broken_links_page_spec.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.describe "Broken links page" do
+  it_behaves_like "redirects non-GDS Editors to services page", "/"
```
leenagupte, line 2: Ah, this is where one of the shared examples is being called. It does read nicer than using `include_examples`

app/controllers/links_controller.rb
```diff
@@ -1,6 +1,9 @@
 class LinksController < ApplicationController
   before_action :load_dependencies, only: %i[edit update destroy]
   before_action :set_back_url_before_post_request, only: %i[edit update destroy]
+  before_action :check_permissions, except: %i[edit update destroy]
```
leenagupte: Should `destroy` be included here? Isn't this saying don't check that it's a GDS Editor when performing a destroy, when it should be the other way around? Am I reading it wrong?

I think this would be easier to follow if it matched the explanation in the comment message,

i.e.
```ruby
before_action :check_permissions, only: %i[index destroy]
```

app/controllers/links_controller.rb
```diff
@@ -33,6 +36,8 @@ def update
   end
 
   def destroy
+    raise GDS::SSO::PermissionDeniedError, "You do not have permission to delete this link" unless gds_editor?
```
leenagupte: Hmmm. So, `destroy` is being explicitly checked? I think I would be nicer to stick to one pattern of checking permissions.

spec/requests/edit_link_page_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        put path, params: { url: }
+
+        expect(link.reload.url).to eq(url)
+      end
+    end
+
+    context "as a department user from the owning department" do
+      before { login_as_department_user(organisation_slug: owning_department) }
+
+      it "returns 403 Forbidden" do
```
leenagupte: Should this test be `it returns 302 Found`?

spec/requests/council_page_spec.rb
```diff
@@ -0,0 +1,79 @@
+RSpec.describe "Council page" do
```
leenagupte, line 1: A test for `get "/bad_homepage_url_status.csv"` should be included somewhere.

app/controllers/services_controller.rb
```diff
@@ -3,10 +3,13 @@ class ServicesController < ApplicationController
...
   before_action :set_service, except: :index
+  before_action :check_permissions, except: :index
+
+  helper_method :org_name_for_current_user
 
   def index
-    @services = Service.enabled.order(broken_link_count: :desc)
-    raise "Missing Data" if @services.empty?
+    @services = services_for_user(current_user).enabled.order(broken_link_count: :desc)
+    # raise "Missing Data" if @services.empty?
```
leenagupte: Should this still be commented out?

app/controllers/services_controller.rb
```diff
@@ -18,6 +21,22 @@ def show
     @breadcrumbs = service_breadcrumbs(@service)
   end
 
+  def update_owner_form
+    raise GDS::SSO::PermissionDeniedError, "You do not have permission to view this page" unless gds_editor?
```
leenagupte: I think to fit the pattern of the other controllers, permissions for this action should be checked in the `before action`, same for `update_owner`

leenagupte: This is a new form? It and it's related bits (view, routes, etc) should really be split out into a separate commit.

leenagupte (commented):
This is looking really good.

I especially like the use of system and request tests.

I've left a few inline comments, but there isn't anything major.

One thing I have noticed though, that Local links manager doesn't have a way to get users back to Sign-on to choose another app, e.g. a departmental publisher wanting to go into Whitehall.

See the "switch app" menu option in Whitehall:
Screenshot 2024-08-13 at 14 13 54

Compared to Local links manager:
Screenshot 2024-08-13 at 14 14 11

If we are opening this up to non-GDS people, we probably need to add something so that they don't get trapped.

app/controllers/services_controller.rb
KludgeKML: No, it should be got rid of entirely (same with the test) - basically, it's perfectly legit for there to be no services returned to a given user now, and if the services all vanish entirely this shouldn't be the way we find out.

leenagupte: 👍  so it just needs to be deleted.

KludgeKML: Done.

app/controllers/services_controller.rb
KludgeKML: I've updated the permissions to match the new pattern.

spec/support/authentication.rb
KludgeKML: Now removed, and this support file is simplified a bit.

app/views/layouts/application.html.erb
KludgeKML: Have done something similar.

spec/support/authentication.rb
KludgeKML: Shared examples now brought in the first time they're used.

app/controllers/links_controller.rb
KludgeKML: I've made three more standard permission checking methods in the concern, so they're all handled as before actions now, and the names are consistent between controllers.

app/controllers/links_controller.rb
KludgeKML: (See above)

spec/requests/edit_link_page_spec.rb
KludgeKML: Thanks! fixed.

spec/requests/council_page_spec.rb
KludgeKML: Added (in a different file, because it's on this controller but isn't a council page per se)

app/presenters/links_table_presenter.rb
KludgeKML: Now removed in the same commit as the test

spec/features/services/service_show_spec.rb
KludgeKML: Now test and code removed in same commit.

spec/controllers/services_controller_spec.rb
KludgeKML: Now removed.

app/controllers/local_authorities_controller.rb
```diff
@@ -3,6 +3,8 @@ class LocalAuthoritiesController < ApplicationController
   include LinkImporterUtils
 
   before_action :set_authority, except: %i[index bad_homepage_url_and_status_csv]
+  before_action :redirect_unless_gds_editor, except: %i[update download_links_csv upload_links_csv]
+  before_action :forbid_unless_gds_editor, only: %i[update download_links_csv upload_links_csv]
```
leenagupte: I find this a real brain teaser, probably because the list of links is the same in both actions. I get there, but it takes a while.

I think that as `bad_homepage_url_and_status_csv` is a file download, it should probably be forbidden if the user is not a `gds_editor` rather than a redirect.

spec/requests/bad_homepage_csv_spec.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.describe "Bad homepage CSV" do
+  it_behaves_like "redirects non-GDS Editors to services page", "/bad_homepage_url_status.csv"
```
leenagupte: Thanks for adding this. 🤩

spec/requests/services_page_spec.rb
```diff
@@ -0,0 +1,16 @@
+RSpec.describe "Services page" do
```
leenagupte, line 1: Should the other routes have tests too? i.e. `download_links_csv` and `upload_links_csv`.

app/controllers/services_controller.rb
```diff
@@ -3,7 +3,8 @@ class ServicesController < ApplicationController
   include LinkImporterUtils
 
   before_action :set_service, except: :index
-  before_action :forbid_unless_permission, except: %i[index]
+  before_action :forbid_unless_permission, except: %i[index update_owner_form update_owner]
+  before_action :forbid_unless_gds_editor, only: %i[update_owner_form update_owner]
```
leenagupte: I'm still finding these permissions confusing. Using "except" and "only" is as confusing as reading "unless, else" for me, especially as the routes in both actions are almost the same.

So if I'm reading it correctly, everyone can see `index`, `update_owner_form` and `update_owner`? 
But the exception for `update_owner_form` and `update_owner` are overwritten on the next line and restricted to `gds_editor` only?

Does `update_owner_form` and `update_owner` need to be covered by `forbid_unless_permission` if they're already covered by  `forbid_unless_gds_editor`?

I for the controllers that the departments can access, I think it's only the `create`, `update` and `destroy` routes that need explicit permissions? The "GET" routes don't need to be restricted as everything else is restricted to only showing the data for the users organisation.

So perhaps all that's needed is the `forbid_unless_gds_editor` action?

```ruby
before_action :forbid_unless_gds_editor, only: %i[update_owner_form update_owner]
```

spec/requests/services_page_spec.rb
```diff
@@ -13,4 +13,60 @@
...
+      it "returns 200 OK" do
+        patch "/services/aardvark-wardens/update-owner", params: { service: { organisation_slugs: "new-owner" } }
+
+        expect(response).to redirect_to("/services/aardvark-wardens?filter=broken_links")
+      end
+
+      it "updates the organisation slugs" do
+        patch "/services/aardvark-wardens/update-owner", params: { service: { organisation_slugs: "new-owner" } }
+
+        expect(service.reload.organisation_slugs).to eq(%w[new-owner])
```
leenagupte, line 122: I think a test is needed for adding multiple owners. Perhaps this one could be updated?

leenagupte (commented):
Thanks for updating how permissions are checked so that they follow a consistent pattern 🎉 

I've resolved all the comments from the previous review. Thanks for fixing.🥇 

The only thing that's still a bit of a brain twister is the logic for permissions in the controllers. I've tried to think of ways to simplify the logic in inline comments, but it's tricky. I think it might be easier to follow if permissions are specifically applied to controller actions. i.e. actions a, b, and c require "gds_editor", d, e and f require "gds_editor" or "organisation", and if the action doesn't require any then, don't mention it. It might make the code a bit more verbose, but there wouldn't be any need to figure out what needs what.

I've also added a few other minor comments inline.

spec/requests/services_page_spec.rb
KludgeKML: Added

spec/requests/services_page_spec.rb
KludgeKML: Added

spec/requests/bad_homepage_csv_spec.rb
KludgeKML: (Marking resolved as positive)

app/controllers/local_authorities_controller.rb
KludgeKML: What I've done is just been explicit about everything. There's an argument that we should do:
```
before_action blah_permission, only: [ etc]
before_action blah_permission 2, only: [ etc_a etc_b]
before_action blah_permission 3, except [listed above]
```

...because it means that if someone adds an action it always gets the final permission. But we don't add actions that frequently, so it might be better to only use `only` and always just list out all actions they apply to.

(I've also fixed the download to make it forbidden)

app/controllers/services_controller.rb
KludgeKML: As above, I've made this so it just uses `only:`

app/controllers/local_authorities_controller.rb
leenagupte: I think perhaps this was made more confusing by having "unless" in the action name. Things being truthy can make the logic easier to follow.
i.e. rather than `forbid_unless_gds_editor` perhaps it could be named `allow_if_gds_editor`. Then even if an `except` was used there'd only be one negative in the logic. I think it was the double negatives that were causing the brain twisting.

leenagupte (approved):
Thanks for updating the controller permissions, I find them very easy to follow now. 🤩 
Thank you bearing with me and making the changes and for adding the extra tests.
Everything looks great👍🥇

PR #1369, opened 2024-08-05
Usability Improvements

KludgeKML:
Usability improvements to tables (moving edit links to a dedicated column at the end, removing additional links and renaming column), and to the local authorities upload links page (errors now appear in a flash, rather than being swallowed and redirected to the local authority page)

#### Before (Multiple confusing calls to action)
Screenshot 2024-08-06 at 09 55 42

#### After (Single call to action link, page title column renamed)
Screenshot 2024-08-06 at 09 55 48

#### Before (Call to action link hidden in authority name )
Screenshot 2024-08-06 at 09 55 19

#### After (Single call to action link)
Screenshot 2024-08-06 at 09 55 30

### Flashes on local authority uploads:
Screenshot 2024-08-06 at 09 52 25

### DIFF
--- a/app/presenters/links_table_presenter.rb
+++ b/app/presenters/links_table_presenter.rb
@@ -47,7 +47,6 @@ class LinksTablePresenter
       },
       {
         text: "",
-        format: "numeric",
       },
     ]
 
--- a/app/presenters/links_table_presenter.rb
+++ b/app/presenters/links_table_presenter.rb
@@ -46,7 +46,7 @@ class LinksTablePresenter
         text: "Status",
       },
       {
-        text: "",
+        text: "<span class=\"govuk-visually-hidden\">Edit</span>".html_safe,
       },
     ]
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,6 +8,7 @@ $govuk-page-width: 1140px;
 @import "govuk_publishing_components/components/breadcrumbs";
 @import "govuk_publishing_components/components/button";
 @import "govuk_publishing_components/components/checkboxes";
+@import "govuk_publishing_components/components/error-alert";
 @import "govuk_publishing_components/components/file-upload";
 @import "govuk_publishing_components/components/heading";
 @import "govuk_publishing_components/components/input";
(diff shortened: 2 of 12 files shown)


andysellick (commented):
Code changes look good as far as I can understand them 😁 

I'm more wondering about the accessibility of the changes. Would it be worth getting Calum to take a look at them? Some specific thoughts:

- links that use repeated text (i.e. 'Edit') aren't accessible because screen reader users don't immediately know what they will edit, normally you'd either make the text more explicit ('Edit _name of thing_') or use a hidden span e.g. `<a href="#">Edit <span class="govuk-visually-hidden">name of thing</span></a>`
- I'm not sure about having edit links on the far right with no column heading, it's possible that people with a high zoom level (because this is a table) might have them go off the edge of the screen, and similar concern for users using a screen magnifier. It might be better to have that column immediately follow the name of the thing column

KludgeKML:
Hmm! Interesting question. We're definitely following a pattern here used in other apps, but yes, looks like there's a known solution for this - we add the visually hidden span to the empty column name, and we add it to the edit links. I'll make those changes.

KludgeKML:
(High zoom level won't push a table off the edge of the page, it'll reflow. Screen magnifier is different, but the state of the art is to put things on the right, so I think it's probably acceptable)

app/presenters/links_table_presenter.rb
```diff
@@ -46,8 +46,7 @@ def headers
         text: "Status",
       },
       {
-        text: "",
-        format: "numeric",
+        text: "<span class=\"govuk-visually-hidden\">Edit</span>".html_safe,
```
andysellick, line 49: The use of a visually hidden span looks correct in this place: https://github.com/alphagov/local-links-manager/pull/1369/files#diff-0c9c05c90dea6bde81d4427841d705abb0c2397c3ad7c9fc763f0d4b84e06a7cR17 but can you explain what this visually hidden `Edit` span does?

KludgeKML: This is the column header. The visual style idiom for these tables is that we don't show a column heading here (it would either be generic/not helpful like "Action" or it would just repeat the visual link, like "Edit"), so we don't want anything visible, but a blank column heading isn't helpful for screen readers, so we add a visually hidden column header. This appears consistent with how (eg) manuals publisher handles this.

andysellick (approved):

PR #1432, opened 2024-10-22
Configure hosts for application

KludgeKML:
- the healthcheck endpoints are requested by IP, not domain, so we need to specifically exclude them from the protection.
- internal API hostname is also necessary

https://trello.com/c/6lU4aUGt/366-configure-hosts-for-application, [Jira issue PNP-8538](https://gov-uk.atlassian.net/browse/PNP-8538)

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -80,10 +80,11 @@ Rails.application.configure do
   config.active_record.dump_schema_after_migration = false
 
   # Enable DNS rebinding protection and other `Host` header attacks.
-  # config.hosts = [
-  #   "example.com",     # Allow requests from example.com
-  #   /.*\.example\.com/ # Allow requests from subdomains like `www.example.com`
-  # ]
+  config.hosts = [
+    /local-links-manager\..*\.gov.uk$/,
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
This fixed worked for a similar issue in Release (https://github.com/alphagov/release/pull/1589) but did not work for a similar issue in Whitehall / Link Checker API: https://github.com/alphagov/whitehall/pull/9542

Sharing just in case the latter trips you up like it did us. We have a [card](https://trello.com/c/VvHwUqD8/3022-investigate-whitehall-403-responses-to-link-checker-api-callbacks) to revisit the latter fix.

KludgeKML:
Ah, interesting! Thanks @ChrisBAshton. I note that if I try to deploy this to integration it appears to fail on the healthcheck, despite that being explicitly excluded from this. Is that similar to what happened on Whitehall?

ChrisBAshton:
Latest PR @KludgeKML : https://github.com/alphagov/whitehall/pull/9560

leenagupte (approved):
Looks ok to me 🤞

